Issue #11, opened 2024-01-29
Add start_time and end_time in listjobs.json


wvengen:
For running jobs returned by the listjobs endpoint, add `start_time`.
For finished jobs returned by the listjobs endpoint, add `start_time` and `end_time`.

See https://scrapyd.readthedocs.io/en/latest/api.html#listjobs-json for an example output:
```json
{
    "status": "ok",
    "pending": [
        {
            "project": "myproject", "spider": "spider1",
            "id": "78391cc0fcaf11e1b0090800272a6d06"
        }
    ],
    "running": [
        {
            "id": "422e608f9f28cef127b3d5ef93fe9399",
            "project": "myproject", "spider": "spider2",
            "start_time": "2012-09-12 10:14:03.594664"
        }
    ],
    "finished": [
        {
            "id": "2f16646cfcaf11e1b0090800272a6d06",
            "project": "myproject", "spider": "spider3",
            "start_time": "2012-09-12 10:14:03.594664",
            "end_time": "2012-09-12 10:24:03.594664",
            "log_url": "/logs/myproject/spider3/2f16646cfcaf11e1b0090800272a6d06.log",
            "items_url": "/items/myproject/spider3/2f16646cfcaf11e1b0090800272a6d06.jl"
        }
    ]
}
```

(we'll omit `items_url`, and `log_url` has #12)

wvengen:
Note that we'd want to get these dates from the container info.
For Docker, try `docker inspect <jobname>` and you'll see `StartedAt` and `FinishedAt` in the `State`.
For Kubernetes, try `kubectl describe job <jobname>` and you'll see _Start Time_ and _Completed At_.

PR #17, opened 2024-02-09
Added start_time and end_time.

Tim-pijnacker:
Improved functionality for _parse_job, such that it now also can return start_time and end_time.

This is only for docker!

Implements #11.

### DIFF
--- a/scrapyd_k8s/launcher/docker.py
+++ b/scrapyd_k8s/launcher/docker.py
@@ -1,5 +1,7 @@
 from datetime import datetime
+
 import docker
+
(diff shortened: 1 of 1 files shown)


wvengen (commented):
Great you found where to get it, and how to process it.
Now you don't really need an extra APIClient, afaik - you already have a connection!
What about `c.attrs['State']`?

wvengen:
p.s. awesome you're having a PR sent in :rocket:

wvengen:
Timeout, closing. Can be picked up by someone else, who will find this PR in the issue for some inspiration.

PR #20, opened 2024-02-12
Added authorization

BoudewijnZwart:
Added optional authorization to the Flask app before every request. Uncomment the fields username and password in the config to enable. The authorization uses Oauth headers.

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,7 +104,7 @@ docker pull ghcr.io/q-m/scrapyd-k8s-spider-example
 
 With `scrapyd-k8s` running and setup, you can access it. Here we assume that
 it listens on `localhost:6800` (for Kubernetes, you would use
-the service name `scrapyd-k8s:6800` instead). If authorization is enabled, you need to send the authorization headers with every request. To use the curl commands below you can just add the option `-u "username:password"` to the curl commands. The username and password can be found (and enabled/disabled) in the config file.
+the service name `scrapyd-k8s:6800` instead). 
 
 ```sh
 curl http://localhost:6800/daemonstatus.json
--- a/README.md
+++ b/README.md
@@ -104,7 +104,7 @@ docker pull ghcr.io/q-m/scrapyd-k8s-spider-example
 
 With `scrapyd-k8s` running and setup, you can access it. Here we assume that
 it listens on `localhost:6800` (for Kubernetes, you would use
-the service name `scrapyd-k8s:6800` instead). 
+the service name `scrapyd-k8s:6800` instead).
 
 ```sh
 curl http://localhost:6800/daemonstatus.json
--- a/README.md
+++ b/README.md
@@ -248,6 +248,7 @@ If you want to delete a project, remove it from the configuration file.
 * `max_proc`     - _(implementation pending)_, if unset or `0` it will use the number of nodes in the cluster, defaults to `0` ([➽](https://scrapyd.readthedocs.io/en/latest/config.html#max-proc))
 * `repository`   - Python class for accessing the image repository, defaults to `scrapyd_k8s.repository.Remote`
 * `launcher`     - Python class for managing jobs on the cluster, defaults to `scrapyd_k8s.launcher.K8s`
+* `username` and `password` - Set both these fields to non-empty to enable basic authentication.
 
 The Docker and Kubernetes launchers have their own additional options.
 
--- a/README.md
+++ b/README.md
@@ -248,7 +248,8 @@ If you want to delete a project, remove it from the configuration file.
 * `max_proc`     - _(implementation pending)_, if unset or `0` it will use the number of nodes in the cluster, defaults to `0` ([➽](https://scrapyd.readthedocs.io/en/latest/config.html#max-proc))
 * `repository`   - Python class for accessing the image repository, defaults to `scrapyd_k8s.repository.Remote`
 * `launcher`     - Python class for managing jobs on the cluster, defaults to `scrapyd_k8s.launcher.K8s`
-* `username` and `password` - Set both these fields to non-empty to enable basic authentication.
+* `username`     - Set this and `password` to enable basic authentication ([➽](https://scrapyd.readthedocs.io/en/latest/config.html#username))
+* `password`     - Set this and `username` to enable basic authentication ([➽](https://scrapyd.readthedocs.io/en/latest/config.html#password))
 
 The Docker and Kubernetes launchers have their own additional options.
 
--- a/README.md
+++ b/README.md
@@ -104,7 +104,7 @@ docker pull ghcr.io/q-m/scrapyd-k8s-spider-example
 
 With `scrapyd-k8s` running and setup, you can access it. Here we assume that
 it listens on `localhost:6800` (for Kubernetes, you would use
-the service name `scrapyd-k8s:6800` instead).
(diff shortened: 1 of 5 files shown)


wvengen:
p.s. Thanks for the PR!

scrapyd_k8s/api.py
```diff
@@ -11,6 +12,16 @@
 config = Config()
 repository = (config.repository_cls())(config)
 launcher = (config.launcher_cls())(config)
+scrapyd_config = config.scrapyd()
+
+config_username = scrapyd_config.get("username")
+config_password = scrapyd_config.get("password")
+
+if not (config_username is None or config_password is None):
```
wvengen, line 20: perhaps rewrite as `if config_username is not None and config_password is not None` for better readability

scrapyd_k8s/api.py
```diff
@@ -11,6 +12,16 @@
...
+    basic_auth = BasicAuth(app)
+    app.config['BASIC_AUTH_USERNAME'] = config_username
+    app.config['BASIC_AUTH_PASSWORD'] = config_password
+    app.config["BASIC_AUTH_FORCE"] = True
```
wvengen, line 24: Can you move this to a function at the end of the file? It is not central to this application, and more like an extra detail you can read about after the rest.
And then here the function can be called to setup authentication for the `app`.

wvengen: I also wonder if the function should be called here or in `run()`, any thoughts?

wvengen (commented):
Ah, this is looking great. Some smaller comments, otherwise good to go! :rocket:

wvengen:
This is still missing documentation in the README too.

README.md
```diff
@@ -104,7 +104,7 @@ docker pull ghcr.io/q-m/scrapyd-k8s-spider-example
 
 With `scrapyd-k8s` running and setup, you can access it. Here we assume that
 it listens on `localhost:6800` (for Kubernetes, you would use
-the service name `scrapyd-k8s:6800` instead).
+the service name `scrapyd-k8s:6800` instead). If authorization is enabled, you need to send the authorization headers with every request. To use the curl commands below you can just add the option `-u "username:password"` to the curl commands. The username and password can be found (and enabled/disabled) in the config file.
```
wvengen, line 107: Please conform to formatting as this happens in the README (wrap lines).

scrapyd_k8s.sample-docker.conf
```diff
@@ -5,6 +5,10 @@
 bind_address = 0.0.0.0
 http_port    = 6800
 
+## uncomment this section to enable authorization
+# username = questionmark
+# password = changeme123
```
wvengen, line 10: * no space between `#` and key, in-line with how this happens in the other sample config
* also add this to the Kubernetes sample config
* please use a more generic username here

scrapyd_k8s/api.py
```diff
@@ -104,7 +106,21 @@ def api_listjobs():
...
-    app.run(host=host, port=port)
\ No newline at end of file
+    host = scrapyd_config.get('bind_address', '127.0.0.1')
+    port = scrapyd_config.get('http_port', '6800')
+    
+    config_username = scrapyd_config.get('username')
+    config_password = scrapyd_config.get('password')
+    
+    if config_username is not None and config_password is not None:
+        basic_auth = enable_authentication(config_username, config_password)
```
wvengen, line 124: unused variable

scrapyd_k8s/api.py
```diff
@@ -104,7 +106,21 @@ def api_listjobs():
 def error(msg, status=200):
     return { 'status': 'error', 'message': msg }, status
 
+def enable_authentication(config_username: str, config_password: str) -> BasicAuth:
```
wvengen, line 109: no type annotations are used elsewhere in this project, so let's not do that here to keep it homogeneous

scrapyd_k8s/api.py
```diff
@@ -104,7 +106,21 @@ def api_listjobs():
...
-    host = config.scrapyd().get('bind_address', '127.0.0.1')
-    port = config.scrapyd().get('http_port', '6800')
-    app.run(host=host, port=port)
\ No newline at end of file
+    host = scrapyd_config.get('bind_address', '127.0.0.1')
+    port = scrapyd_config.get('http_port', '6800')
+    
+    config_username = scrapyd_config.get('username')
+    config_password = scrapyd_config.get('password')
+
```
wvengen, line 122: remote this empty line, as these four lines belong together

scrapyd_k8s/api.py
wvengen, line 109: pass `app` here, don't use the global

wvengen (commented):
Super! Some smaller things, otherwise good to go :+1:

README.md
wvengen: Btw, we don't need to mention this. It is not enabled by default, and if you enable it, you will know what you are doing.

What I meant was: add it to the list of general configuration variables at the bottom of the README.

scrapyd_k8s.sample-docker.conf
```diff
@@ -5,9 +5,9 @@
 bind_address = 0.0.0.0
 http_port    = 6800
 
-## uncomment this section to enable authorization
-# username = questionmark
-# password = changeme123
+## Uncomment this section to enable authorization
+#username    = sampleusername
+#password    = changeme123
```
wvengen, line 11: Please really look at the surrounding style and stick to it (and if you suggest to change the style, discuss it).
In this cases, that would be single `##` also for the comment.
Other comments also end with a dot `.`, so that would be good to use here as well.

Could you use as comment: `# Optional basic http authentication for all endpoints.` ?

wvengen (changes requested):
Almost there!
And don't forget to document the options at the end of the README.

BoudewijnZwart:
> Almost there! And don't forget to document the options at the end of the README.

You mean the options username and password right? So like this? 
image

BoudewijnZwart:
Or with more information about the authorization?

BoudewijnZwart:
Sorry I mean like this, without the periods:
image

wvengen:
Ah yes, something like that :) And it can be a little shorter, e.g. something like "`username` - set this and `password` to enable basic authentication (->)"

(actually, empty is also fine, since the code checks for `None` ;))

BoudewijnZwart:
I took the "non-empty" from the scrapyd docs. But I think "set this and `password` to enable basic authentication" is indeed better, because technically an empty string would also enable authentication.

README.md
```diff
@@ -104,7 +104,7 @@ docker pull ghcr.io/q-m/scrapyd-k8s-spider-example
 
 With `scrapyd-k8s` running and setup, you can access it. Here we assume that
 it listens on `localhost:6800` (for Kubernetes, you would use
-the service name `scrapyd-k8s:6800` instead).
+the service name `scrapyd-k8s:6800` instead).
```
wvengen, line 107: Last final thing: spurious extra space :)

wvengen (changes requested):
Super!
Allmost ready :)

BoudewijnZwart:
Ah good catch.

wvengen (approved):
Hooray! :tada:

PR #27, opened 2024-03-06
Add nodename to response

vlerkin:


scrapyd_k8s/api.py
```diff
@@ -14,6 +14,15 @@
...
 
+# middleware that adds "node_name" to each response if it is a JSON
+@app.after_request
+def after_request(response: Response):
+    if response.is_json:
+        data = response.json
+        data["node_name"] = config.scrapyd().get("node_name", launcher.get_node_name())
+        response.data = jsonify(data).data
+    return response
+
```
wvengen, line 25: Let's move this more near the end of the file, after the routes, because it is more like an extra thing to add, not the main thing.

scrapyd_k8s/launcher/docker.py
```diff
@@ -18,6 +20,9 @@ class Docker:
     def __init__(self, config):
         self._docker = docker.from_env()
 
+    def get_node_name(self):
+        return os.getenv('HOSTNAME')
+
```
wvengen, line 25: Is this always set?

scrapyd_k8s/launcher/k8s.py
```diff
@@ -23,6 +25,9 @@ def __init__(self, config):
         self._k8s = kubernetes.client.CoreV1Api()
         self._k8s_batch = kubernetes.client.BatchV1Api()
 
+    def get_node_name(self):
+        return "%s.%s" % (os.getenv('MY_DEPLOYMENT_NAME'), os.getenv('MY_POD_NAMESPACE'))
+
```
wvengen, line 30: Let's have a fallback here. Someone's deployment may _not_ have these variables set.

test_api.py
```diff
@@ -38,13 +38,20 @@ def test_daemonstatus_ok():
 def test_listprojects_ok():
     response = requests.get(BASE_URL + '/listprojects.json')
     assert_response_ok(response)
-    assert response.json() == { 'status': 'ok', 'projects': AVAIL_PROJECTS }
+
+    json = response.json()
+    assert json['projects'] == AVAIL_PROJECTS
+    assert 'node_name' in json
```
wvengen, line 44: What about setting a `node_name` in the example config, then we can also test its value?

We don't have the infrastructure to test config file variations at the moment.

test_api.py
```diff
@@ -113,14 +132,15 @@ def test_scenario_regular_ok():
...
+
 def test_scenario_regular_ok_without_version():
     scenario_regular({
         'project': RUN_PROJECT, 'spider': RUN_SPIDER,
         'setting': 'STATIC_SLEEP=%d' % STATIC_SLEEP
     })
 
-# TODO test_scenario_cancel_scheduled_ok (needs a way to make sure a job is not running yet)
 
+# TODO test_scenario_cancel_scheduled_ok (needs a way to make sure a job is not running yet)
```
wvengen, line 143: unrelated whitespace change

wvengen (changes requested):
Well done, great! Some small comments, otherwise good to go :+1:

scrapyd_k8s/launcher/k8s.py
vlerkin: What kind of fallback do you want here?

wvengen: Are there any variables that are set by default in Kubernetes pods that could be useful?
If not, then we could just use `default` here.

test_api.py
vlerkin: Maybe let's make it a ticket out of it?

wvengen: Ok, maybe good indeed to not have this in the example config files.

scrapyd_k8s/launcher/k8s.py
wvengen: Thanks for looking this up!
I'm doubting a bit whether we
1. keep it simple stupid (and fallback to something like `default`) - as you can always set the node name in the config
2. return the pod name, so that there is at least something (but returns a new pod name after each restart, which would be polluting systems showing a list of node names, where multiple scrapyd servers are being accessed)
3. remove the suffix from the pod name to return the deployment name - this is a bit of a hack, and may give weird results if someone does not use a deployment, for example

I think the first option is actually the simplest to understand, I think.
Any thoughts?

scrapyd_k8s/launcher/k8s.py
```diff
@@ -26,6 +26,10 @@ def __init__(self, config):
         self._k8s_batch = kubernetes.client.BatchV1Api()
 
     def get_node_name(self):
+        deployment_name = os.getenv('MY_DEPLOYMENT_NAME')
+        pod_name = os.getenv('MY_POD_NAMESPACE')
+        if not deployment_name and not pod_name:
+            return "default"
         return "%s.%s" % (os.getenv('MY_DEPLOYMENT_NAME'), os.getenv('MY_POD_NAMESPACE'))
```
wvengen, line 33: What about?

```python
deployment = os.getenv('MY_DEPLOYMENT_NAME', 'default')
namespace = os.getenv('MY_NAMESPACE')
return ".".join([n for n in [pod_name, deployment_name] if n])
```

Also, looking at this, I wonder if it makes sense to put this in the Kubernetes manifest, and just have `SCRAPYD_K8S_NODE_NAME`, and build this a variable in the Kubernetes manifest.

test_api.py
```diff
@@ -15,20 +15,17 @@
 MAX_WAIT = int(os.getenv('TEST_MAX_WAIT', '6'))
 STATIC_SLEEP = float(os.getenv('TEST_STATIC_SLEEP', '2'))
 
-
```
wvengen, line 18: unrelated whitespace changes

wvengen:
After a CI rerun also the Kubernetes test is fine again :sweat_smile:

scrapyd_k8s/launcher/docker.py
```diff
@@ -21,7 +21,11 @@ def __init__(self, config):
         self._docker = docker.from_env()
 
     def get_node_name(self):
-        return os.getenv('HOSTNAME')
+        hostname = os.getenv('HOSTNAME')
+        if hostname in [None, "", "null"]:
```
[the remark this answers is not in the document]
vlerkin: Most often HOSTNAME exists in linux based systems but it's not part of posix family standard so there can be cases when it's null, this is the source my thoughts are based on:
https://superuser.com/questions/132489/hostname-environment-variable-on-linux#comment1214603_132500

scrapyd_k8s/launcher/docker.py
```diff
@@ -21,7 +21,11 @@ def __init__(self, config):
...
+            hostname = os.popen("hostname").read().strip()
```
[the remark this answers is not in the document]
vlerkin: Do you think it would be better to use another approach here?

scrapyd_k8s/launcher/docker.py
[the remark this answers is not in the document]
wvengen: What brings `HOSTNAME` that `gethostname()` doesn't bring? (Perhaps the ability to override the node name from the environment, but then I'd like it across launcher backends, and not provide that ability here.) So I would just use `gethostname()`. You can already override the node name in the config file.

wvengen (commented):
One pending thing, otherwise good to go!

scrapyd_k8s/launcher/docker.py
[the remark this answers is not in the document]
vlerkin: Yes, true, os.getenv is just a call to "third-party" program while gethostame is part of core if I understand correctly so better just use gethostname and in case it's null then use "default"?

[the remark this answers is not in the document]
wvengen: Well, both are from the core Python library, but `os.getenv()` looks in the environment, and `gethostname()` does a syscall to request the system hostname. The Superuser question shows that `HOSTNAME` may not be a proper environment variable (and if it is may depend on the Linux distribution or other *nix system flavour, in any case it is not part of the POSIX spec), so that you cannot really rely on it.

wvengen (commented):
Thanks!
It's good to always review your own PR. Doing a review of all changes, shows me that there are still unrelated whitespace changes with meaning (e.g. the newlines in the tests do convey something about the grouping - e.g. a comment of a missing test does not belong to the following method, so you can't just remove the empty newline. Linters and autoformatters don't have the final word!)

Could you still review and change this? When the whitespace is correct, the PR can be merged.

vlerkin:
> Thanks! It's good to always review your own PR. Doing a review of all changes, shows me that there are still unrelated whitespace changes with meaning (e.g. the newlines in the tests do convey something about the grouping - e.g. a comment of a missing test does not belong to the following method, so you can't just remove the empty newline. Linters and autoformatters don't have the final word!)
> 
> Could you still review and change this? When the whitespace is correct, the PR can be merged.

I deleted whitespace added by code formatter, if you notice there is still some whitespace that you think should be different - let me know. I am now having my Mac wiped out so I won't be able to access it for some time.

wvengen (approved):
Good enough, thank you for all the changes! :rocket:

PR #30, opened 2024-08-30
Add job start and end time

### LINKED ISSUE #11: Add start_time and end_time in listjobs.json

leewesleyv:
Resolves #11

- [x] Add start time to job data for the Docker launcher
- [x] Add start time to job data for the K8s launcher
- [x] Add end time to job data for the Docker launcher
- [x] Add end time to job data for the K8s launcher
- [x] Add unit tests
- [x] Add documentation


scrapyd_k8s/launcher/docker.py
```diff
@@ -55,36 +60,44 @@ def schedule(self, project, version, spider, job_id, settings, args):
...
-        prevstate = self._docker_to_scrapyd_status(c.status)
-        if c.status == 'created' or c.status == 'scheduled':
-            c.remove()
-        elif c.status == 'running':
-            c.kill(signal='SIG' + signal)
+        prevstate = self._docker_to_scrapyd_status(container.status)
+        if container.status == 'created' or container.status == 'scheduled':
+            container.remove()
+        elif container.status == 'running':
+            container.kill(signal='SIG' + signal)
```
wvengen, line 71: This is a change in code-style, I'd prefer to keep it as `c`.
Yes, often quite indicative names are often a good idea. But in this case, it is such a common domain object here that I would think it's obvious, and provide an easier to read experience (terser).
And now there are different code styles in the same project / file, which is more confusing.
So unless you have important objections, please revert.

scrapyd_k8s/launcher/docker.py
```diff
@@ -55,36 +60,44 @@ def schedule(self, project, version, spider, job_id, settings, args):
...
-            'project': c.labels.get(self.LABEL_PROJECT),
-            'spider': c.labels.get(self.LABEL_SPIDER)
+    def _parse_job(self, container: Container):
+        state = self._docker_to_scrapyd_status(container.status)
+        job = {
+            'id': container.labels.get(self.LABEL_JOB_ID),
+            'state': state,
+            'project': container.labels.get(self.LABEL_PROJECT),
+            'spider': container.labels.get(self.LABEL_SPIDER),
         }
```
wvengen, line 81: Why moving this out of the dict?

test_api.py
```diff
@@ -159,7 +161,11 @@ def test_scenario_cancel_running_finished_ok():
     # wait until the job has stopped
     listjobs_wait(jobid, 'finished')
     jobinfo = assert_listjobs(finished=jobid)
-    assert jobinfo == { 'id': jobid, 'project': RUN_PROJECT, 'spider': RUN_SPIDER, 'state': 'finished' }
+    assert jobinfo['id'] == jobid
+    assert jobinfo['project'] == RUN_PROJECT
+    assert jobinfo['spider'] == RUN_SPIDER
+    assert jobinfo['state'] == 'finished'
+    assert datetime.strptime(jobinfo['start_time'], TIME_FORMAT)
```
wvengen, line 168: I'd rather keep the previous form, so that if anything is added to the dict, you are aware that a test needs changing.

scrapyd_k8s/settings.py
```diff
@@ -0,0 +1 @@
+TIME_FORMAT = '%Y-%m-%d %H:%M:%S.%f'
```
wvengen, line 1: I'm not really sure if `settings` is a good name here, I would think this has to do with user-configurable settings.
Maybe `const` or `util`?
Very good to centralize this somewhere!

scrapyd_k8s/launcher/docker.py
```diff
@@ -55,36 +60,44 @@ def schedule(self, project, version, spider, job_id, settings, args):
...
+            job['start_time'] = datetime.fromisoformat(container.attrs['State']['StartedAt']).strftime(TIME_FORMAT)
+        if state == 'finished':
+            job['end_time'] = datetime.fromisoformat(container.attrs['State']['FinishedAt']).strftime(TIME_FORMAT)
+
+        return job
+
+    def _get_container(self, project_id: str, job_id: str) -> Optional[Container]:
+        filters = {'label': self.LABEL_JOB_ID + '=' + job_id}
+        containers = self._docker.containers.list(all=True, filters=filters)
+        if not containers:
```
wvengen, line 93: Here it is actually more readable to use a plural.
Still, perhaps it's good to do it at more places than just here, so I'd vote to keep it as it is for now.

scrapyd_k8s/launcher/k8s.py
```diff
@@ -170,13 +184,12 @@ def _k8s_job_name(self, project, job_id):
 
     def _k8s_kill(self, pod_name, signal):
         # exec needs stream, which modifies client, so use separate instance
-        k8s = kubernetes.client.CoreV1Api()
-        resp = kubernetes.stream.stream(
-            k8s.connect_get_namespaced_pod_exec,
+        kubernetes.stream.stream(
```
wvengen, line 187: please keep `resp` here, to remind the reader that it has a result that we may actually want to do something with

wvengen (commented):
Nice!
Great to see some cleanup too (though some comments).
I still need to get used to the idea of introducing typing here. Maybe discuss this first at a standup, and then perhaps introduce later?

scrapyd_k8s/launcher/docker.py
leewesleyv: We move the `state` variable out of the dict so that we can use the variable later on to make an evaluation (for adding start/end time). If that is what you were referring to.

scrapyd_k8s/settings.py
leewesleyv: I was struggling with the location too. Since there is not yet a common place to put these things. We already have a `utils.py`, but not sure if this would be a right fit there?

scrapyd_k8s/launcher/docker.py
wvengen: Ah, sorry, I missed that, thanks for explaining!

test_api.py
leewesleyv: I will update this. Would need to look a bit more into mocking the `start_time` and `end_time` of the jobs during tests, since a dict comparison will require the exact `start_time` and `end_time` of the container/pods. Or maybe a different solution.

scrapyd_k8s/launcher/docker.py
leewesleyv: Let's keep it consistent in the project. If we are going the change the code style, we should indeed do it for the entire project. But that should not be part of this PR!

test_api.py
wvengen: Ah, that makes sense.
What about removing the time fields from the dict under test, and testing these separately? That would both allow us to test the full response, and allow proper time comparison.
Or use a time-freezing testing library (there must exist one).

scrapyd_k8s/settings.py
wvengen: I think that makes sense, let's do that.
We _could_ also make it into a function that does the `strftime`.

test_api.py
leewesleyv: I tried using `freezegun` to set a specific time during testing, but since we are talking to the kubernetes/docker APIs and not using the datetime module, this solution did not work. I will do a bit more digging. The first solution should also work though, but it might be better to see if there are alternative solutions.

wvengen: Ah, that totally makes sense! I'm not for mocking the result in the component under test, so let's fix this in the testing part.

scrapyd_k8s/launcher/docker.py
```diff
@@ -1,9 +1,9 @@
 import re
-import os
 import socket
 
 import docker
-from ..utils import native_stringify_dict
+
+from scrapyd_k8s.utils import format_iso_date_string, native_stringify_dict
```
wvengen, line 6: I'm not fully sure if moving to absolute imports has unintended consequences.

wvengen (commented):
Looking good, thanks!
I see you've added the first unit test. Great!
The API test is not a regular test, as it is meant to run against a running instance. I'm not really sure if it fits well into the general pytest infrastructure like this. What do you think?

I'd split the API tests (which can be run completely separate from scrapyd-k8s, and could even be run against a different implementation, iirc), and the unit tests, into separate CI jobs.

scrapyd_k8s/launcher/docker.py
wvengen: e.g. running the daemon from a different PWD (not sure if that works anyway), it might e.g. impact the Docker image

scrapyd_k8s/tests/integration/test_api.py
```diff
@@ -1,9 +1,12 @@
 #!/usr/bin/env python3
+from datetime import datetime
 import os
-import pytest
 import requests
 import time
 
+from scrapyd_k8s.api import launcher
+from scrapyd_k8s.launcher.k8s import K8s
+from scrapyd_k8s.utils import TIME_FORMAT
```
wvengen, line 10: Ok, you really cannot import anything from the project's code here. Please keep this an integration test that tests the API. Even `TIME_FORMAT` is something to hardcode here - if it changes in the code, this is an API breakage, so we want to catch this in the tests.

scrapyd_k8s/tests/integration/test_api.py
```diff
@@ -159,6 +163,16 @@ def test_scenario_cancel_running_finished_ok():
...
     jobinfo = assert_listjobs(finished=jobid)
+    start_time, end_time = jobinfo.pop('start_time'), jobinfo.pop('end_time')
+    assert start_time and datetime.strptime(start_time, TIME_FORMAT)
+
+    # Only when a k8s job has succeeded will it have a completion time. Here we cancel
+    # the job and thus we expect the end_time to undefined.
+    if isinstance(launcher, K8s):
+        assert not end_time
+    else:
+        assert end_time and datetime.strptime(end_time, TIME_FORMAT)
```
wvengen, line 174: I understand that you are tempted to check the `launcher` here.
This is an integration test and it needs to have a well-defined behaviour. No ifs like this here.
If you want to control the job, then we need to either setup the environment to do this, or find another way. E.g. add a spider setting to the example spider that lets it return with a non-success exit code, or raise an error. Or add something to the scrapyd-k8s to schedule a job without starting it.

Do _not_ test if running on k8s or not - if you _really_ need to distinguish behaviour, make it explicit, e.g. add an evironment variable to the test that exactly tells what you expect in the test (e.g. `TEST_END_TIME_MISSING=1`).

It looks like behaviour on Docker vs. Kubernetes different. Why? I don't think we'd want this.

scrapyd_k8s/tests/integration/test_api.py
wvengen: I think we will need to document this in a comment on top of the file, that this test is meant to be run against a running instance of scrapyd-k8s, and is a _standalone_ test (and does and must not reference scrapyd-k8s code). We are testing scrapyd-k8s REST API conformity and behaviour, _not_ the Python code.

scrapyd_k8s/launcher/k8s.py
```diff
@@ -122,13 +121,24 @@ def cancel(self, project, job_id, signal):
         return prevstate
 
     def _parse_job(self, job):
-        return {
+        state = self._k8s_job_to_scrapyd_status(job)
+        _job = {
```
wvengen, line 125: Can we come up with a better name here? As a casual reader, I would be confused having two jobs here.

scrapyd_k8s/tests/integration/test_api.py
leewesleyv: Adding the `end_time` will require an entirely different approach. The behaviour is different when using Docker vs. Kubernetes. Reverted the code for now so we can come up with something else. Not sure yet on how to approach this.

wvengen: Ok, let's discuss it in the issue!
We could also include it just for Kubernetes, but then I'd like to document it, and know what would be needed (and probably it could make sense to explain it in the Docker launcher code).

README.md
```diff
@@ -149,9 +149,9 @@ curl http://localhost:6800/listjobs.json
 ```
 ```json
 {
-  "finished":[],
+  "finished":[{"id":"fbjffo6tz5al5qjft0zrv0t7b0vzccxx","project":"example","spider":"static","state":"finished", "start_time":"2012-09-12 10:10:00.000000", "end_time":"2012-09-12 10:20:00.000000"}], // End time will only be available for the k8s launcher (when the job is finished and not cancelled), for docker it will always be null
```
leewesleyv, line 152: @wvengen do you think this is sufficient? Or should we also mention something in the **API** > **listjobs.json** section in the readme?

wvengen: Ouch, this is ugly. This output shows what you would get when running this. You don't see this comment.

I would expect this either in a section describing current features and limitations, or perhaps in our case, in the list of endpoints supported. There you can indicate what job status does not support.

leewesleyv: Added a section for limitation to the listjobs sections in the readme.

wvengen:
Can we make it a bit terser? (a separate heading breaks the flow of reading through the endpoints; this is just a detail you're only interested in if you want to know more about this endpoint)

```diff
--- a/README.md
+++ b/README.md
@@ -228,10 +228,7 @@ Lists spiders from the spider image's `org.scrapy.spiders` label.
 ### `listjobs.json` ([➽](https://scrapyd.readthedocs.io/en/latest/api.html#listjobs-json))
 
 Lists current jobs by looking at Docker containers or Kubernetes jobs.
-
-#### Limitations
-
-* **End time**; The job's end time will be populated only for the Kubernetes (k8s) launcher, provided the job finishes successfully and is not canceled. For Docker, this value will always be null.
+Note that `end_time` is not yet supported for Docker.
 
 ### ~~`delversion.json`~~ ([➽](https://scrapyd.readthedocs.io/en/latest/api.html#delversion-json))
 
```

wvengen:
Otherwise, good to go!

leewesleyv:
@wvengen Sorry, I did not see your suggestion. Do you think the change I've made is OK? Or should I change it to the suggestion?

wvengen:
Yes, please use the diff I posted. Thanks!

README.md
```diff
@@ -228,6 +228,7 @@ Lists spiders from the spider image's `org.scrapy.spiders` label.
 ### `listjobs.json` ([➽](https://scrapyd.readthedocs.io/en/latest/api.html#listjobs-json))
 
 Lists current jobs by looking at Docker containers or Kubernetes jobs.
++Note that `end_time` is not yet supported for Docker.
```
wvengen, line 231: spurious `+` here

wvengen:
Super, thanks!

PR #31, opened 2024-09-10
resource watcher, log watcher, libcloud, object_storage

vlerkin:
As a separate thread, resource watcher was added. When a new pod with specific labels (job labels) are aded to the cluster and are in a running state, another watcher starts to follow it's logs and write it to a file stored on a managing pod with scrapyd. Then all files are uploaded to an object storage of a preferred provider, which is implemented via Apache Libcloud library. Persistent Volume and PVC are commented out in the Kubernetes yaml file and were not tested in the cloud, but it can be used to add them to the managing pod to store log files in a persistent volume.


kubernetes.yaml
```diff
@@ -80,6 +85,15 @@ data:
...
     launcher     = scrapyd_k8s.launcher.K8s
+    
+    num_lines_to_check = 2
+    
+    storage_provider = s3
+    
+    aws_region            = <aws_region>
+    aws_access_key_id     = <aws_access_key_id>
+    aws_secret_access_key = <aws_secret_access_key>
+    aws_bucket_name       = <aws_bucket_name>
```
wvengen, line 96: Good to see something in the configuration file. I think this warrants its own section, maybe `logging` but that could also refer to the daemon's logging settings, so perhaps something like `[logwatcher]` or `[logging.jobs]` or something.

kubernetes.yaml
```diff
@@ -80,6 +85,15 @@ data:
...
     repository   = scrapyd_k8s.repository.Remote
     launcher     = scrapyd_k8s.launcher.K8s
+    
+    num_lines_to_check = 2
+    
+    storage_provider = s3
+    
+    aws_region            = <aws_region>
+    aws_access_key_id     = <aws_access_key_id>
+    aws_secret_access_key = <aws_secret_access_key>
```
wvengen, line 95: It's bad practice to have secrets in a configmap. For kubernetes, we could mention a secret name (but then the daemon needs to have access to the secret), or perhaps better using environment variables, and then the manifest can set these from a secret (giving the flexibility to the deployment configuration). For Docker, we need to use the environment anyway.

kubernetes.yaml
```diff
@@ -106,6 +120,34 @@ metadata:
...
+#kind: PersistentVolumeClaim
+#metadata:
+#  name: efs-claim
+#spec:
+#  accessModes:
+#    - ReadWriteOnce
+#  storageClassName: ""
+#  resources:
+#    requests:
+#      storage: 5Gi
```
wvengen, line 150: Good you added these (commented out) to the manifest. Please reduce is to the minimum needed to run on minikube (which is used on CI). Even though this is commented out, I think it's good to tune to that use-case, then people can adapt to their own cloud provider (so no mention of efs).

kubernetes.yaml
wvengen: Or if we want to make this project-specific as well, then we'd perhaps use `[default.logs]` and `[<project>.logs]` or so.

scrapyd_k8s/log_handler.py
```diff
@@ -0,0 +1,54 @@
...
+    try:
+        with open(file_path, 'r') as file:
+            lines = file.readlines()
+        return lines[-num_lines_to_check:]
+    except FileNotFoundError:
+        return []
+
+def concatenate_and_delete_files(main_file_path, temp_file_path):
+    try:
+        subprocess.run(f"cat {temp_file_path} >> {main_file_path}", shell=True, check=True)
```
wvengen, line 16: Is there a reason you're doing this in a subprocess, rather than in Python?

scrapyd_k8s/log_handler.py
```diff
@@ -0,0 +1,54 @@
...
+        os.remove(temp_file_path)
+    except Exception as e:
+        print(f"CONCATENATION ERROR: {e}")
+
+def make_log_filename_for_job(job_name):
+    return os.path.join('/tmp', f"{job_name}_logs.txt")
```
wvengen, line 22: please use `mkstemp` and friends, or rather the Python alternative, to avoid naming race conditions (even though we're probably running in a container). Or perhaps a temporary directory to keep `/tmp` a bit clean. Also, look into standard environment variables to set the temporary directory (like `TMPDIR`, though ideally this would be handled by the Python tempfile-related standard library, not sure).

scrapyd_k8s/object_storage/libcloud_driver.py
```diff
@@ -0,0 +1,46 @@
...
+        self._container_name = config.scrapyd().get('aws_bucket_name')
+        self._storage_provider = config.scrapyd().get('storage_provider')
+
+        if self._region is None or self._access_key_id is None or self._secret_access_key is None:
+            raise ValueError('AWS credentials not set')
+        if self._container_name is None:
+            raise ValueError('Bucket name not set')
+
+        driver_class = get_driver(self._storage_provider)
+        self.driver = driver_class(self._access_key_id, self._secret_access_key, region=self._region)
```
wvengen, line 21: Ideally, we'd have some kind of config that is passed to libcloud, so we don't need to put all known configuration keys for libcloud here. I'll look into it when I have some time.

scrapyd_k8s/object_storage/libcloud_driver.py
```diff
@@ -0,0 +1,46 @@
...
+        try:
+            object_from_storage = self.driver.get_object(container_name=self._container_name, object_name=object_name)
+            if object_from_storage is not None:
+                return True
+        except ObjectDoesNotExistError:
+            return False
+        except ContainerDoesNotExistError as e:
+            print(f"Container with the name '{self._container_name}' does not exist in the cloud storage")
+        except InvalidContainerNameError as e:
+            print(f"Invalid container name '{self._container_name}'")
```
wvengen, line 44: We need to think about error handling a bit. Fine to do that later!
(and if we'd decide on logging and not doing anything more, then we may want to use Python's logging framework)

kubernetes.yaml
```diff
@@ -51,6 +51,8 @@ spec:
               mountPath: /opt/app/scrapyd_k8s.conf
               readOnly: true
               subPath: scrapyd_k8s.conf
+            #- name: persistent-storage
+            #  mountPath: /data
```
wvengen, line 55: Let's name this according to what the persistent storage is used for, so I guess that would be something logging-related, like `logwatch`, `joblogs`, or whatever we decide on.

wvengen (commented):
Wow, very very nice to see a not-gigantic, mostly understandable piece of code already! I'll need to read it a few times to fully grasp what is happening, and what we'd want for the project, but I've already added some comments that may help shape the direction.

requirements.txt
```diff
@@ -3,3 +3,6 @@ kubernetes>=27.2.0 # introduction of suspend in jobspec
 flask>=2.0.0
 natsort>=8.0.0
 Flask-BasicAuth>=0.2.0
+MarkupSafe>=2.1.5
+apache-libcloud>=3.8.0
+requests>=2.31.0
```
wvengen, line 8: requests is not a direct dependency, why do you need to add this? (and if you need, then it would make sense to document it in the requirements file)

scrapyd_k8s.sample-k8s.conf
```diff
@@ -12,6 +12,19 @@ http_port    = 6800
 # Kubernetes works with a remote repository and the Kubernetes launcher.
 repository   = scrapyd_k8s.repository.Remote
 launcher     = scrapyd_k8s.launcher.K8s
+object_storage_provider  = scrapyd_k8s.object_storage.LibcloudObjectStorage
```
wvengen, line 15: It's strange that you include this line here, but not in the docker config file.
I think we can leave this out this line, as there is only one implementation, and there is a sensible default (great!)

I would say that the job logging feature is an additional option, let's disable it by default, and remove it from the example configfiles. We might add it fully commented out (near the end of the file), to show how it could be configured, but I'd rather put that in documentation for now.

scrapyd_k8s/api.py
```diff
@@ -148,5 +189,11 @@ def run():
     if config_username is not None and config_password is not None:
         enable_authentication(app, config_username, config_password)
 
+    pod_watcher_thread = threading.Thread(
+        target=watch_pods,
+        kwargs={'namespace': scrapyd_config.get('namespace', 'default')}
```
wvengen, line 194: I think there are now multiple places where we could get the namespace, and provide a default. I think I might add a `scrapyd_config` method to get the namespace, and provide the default there, so that this is consistent across the places where the namespace is used.

scrapyd_k8s/api.py
```diff
@@ -148,5 +189,11 @@ def run():
...
+    )
+    pod_watcher_thread.start()
+
```
wvengen, line 197: It's maybe a bit early to start about performance-related things, but I think that the job logging feature will not be used by everyone, and adds overhead: for every change in pod state for every pod in the namespace, this Python program examines it to decide whether to do anything with it. So let's only start the pod watcher if we will actually use it (so don't when the job logging is not enabled, or if there is a configuration issue with it that allows scrapyd-k8s to run but the job logging to not be active - if that situation could happen).

scrapyd_k8s/log_handler.py
```diff
@@ -0,0 +1,54 @@
+import os
+from kubernetes import client, watch
+import subprocess
+
+def get_last_n_lines(file_path, num_lines_to_check):
+    lines = []
+    try:
+        with open(file_path, 'r') as file:
+            lines = file.readlines()
+        return lines[-num_lines_to_check:]
```
wvengen, line 10: Performance-related tweak: seek to the end of the file, then read chunks in reverse order, and determine what the last lines are. This can be a significant improvement if the log files are huge. See e.g. [this SO](https://stackoverflow.com/a/23646049/2866660) or [`file_read_backwards`](https://pypi.org/project/file-read-backwards/).

Please keep this possible improvement for the end, let's first get the implementation done! (I needed a place to jot this down)

scrapyd_k8s/api.py
```diff
@@ -137,6 +143,41 @@ def enable_authentication(app, config_username, config_password):
...
+                watcher_threads[thread_name] = threading.Thread(
+                    target=stream_logs,
+                    kwargs={
+                        'job_name': pod.metadata.name,
+                        'namespace': namespace,
+                        'num_lines_to_check': num_lines_to_check
+                    }
+                )
+                watcher_threads[thread_name].start()
+        elif pod.status.phase == 'Succeeded' and pod.metadata.labels.get("org.scrapy.job_id"):
```
wvengen, line 169: What happens if the job errored? Is the phase then also `Succeeded`?
I think we would like to see the logs also when it errored (e.g. the process terminated).

scrapyd_k8s/api.py
```diff
@@ -137,6 +143,41 @@ def enable_authentication(app, config_username, config_password):
...
+            log_filename = make_log_filename_for_job(pod.metadata.name)
+            if os.path.isfile(log_filename):
+                if object_storage_provider.is_local_file_uploaded(log_filename):
+                    print("file already exists")
+                else:
+                    object_storage_provider.upload_file(log_filename)
+            else:
+                print("logfile not found %s", pod.metadata.name)
+        else:
+            print("other type " + event["type"] + " " + pod.metadata.name + " - " + pod.status.phase)
```
wvengen, line 179: Can we move this piece of code to the log handling file? I'd like to keep the `api.py` file as clean as possible. Then the job log handling functionality kind of 'uses' the watch functionality provided by the main api program.

scrapyd_k8s/api.py
wvengen: Combined with the previous comment of moving the functionality that does something with the watched pod info, the code here could ask the job logging code if it needs the pod watcher or not (depending on its configuration).

wvengen (commented):
I've reviewed it a bit more. Still happy that what is actually happening, is not that complex to understand. That is a sign of clear coding, well done!

I think the next steps would be:
- [ ] Move more job logging related code to its own module (incl. objectstorage)
- [ ] Polish `api.py` related to this change
- [ ] Consistently use `joblogs` terminology (or whatever we finally settle on)
- [ ] Find a clean solution for storage configuration
- [ ] Error handling and/or logging
- [ ] The other smaller security- and performance-related things

wvengen:
And regarding naming: I would propose to call this feature `joblogs` (or maybe `job logging`).

I think that ideally we'd move all files related to this feature to a subdirectory `joblogs`, which would include object storage (if we would start using object storage in another component as well, then we would move it to its own subdirector). In that way, all things related to this feature can be found there, and you can easily ignore it if you don't want to dive into this.

scrapyd_k8s/log_handler.py
vlerkin: It seemed like a simple way, but maybe introduces some overhead, I'll use Python with open thing to append a file to an existing one

scrapyd_k8s/api.py
vlerkin: If the job errors, then the status of the pod is "Failed". In our schedule function we aim to 1 completion but have  backoff_limit=0, so if a job fails at least 1 time, the pod status will change (the status of pod is changing to failed if at least 1 container inside has a non-zero exit code).

vlerkin: I will add code to collect logs from failed pods, need to think a bit if it belongs to the main loop with a happy scenario.

scrapyd_k8s/api.py
vlerkin: Please, let me know if you changed it so I can adopt your edits, or if you want I can add it

wvengen: Note that for k8s there already is a namespace setting on the configuration. Sorry, I meant that you could add this method, if you think that's a good idea :) To avoid the default namespace being set in multiple places, please make a method in the configuration class, which also provides the default.

kubernetes.yaml
```diff
@@ -59,6 +61,9 @@ spec:
         - configMap:
             name: scrapyd-k8s-config
           name: scrapyd-k8s-config
+        #- name: joblogs
+        #  persistentVolumeClaim:
+        #    claimName: efs-claim
```
wvengen, line 66: Can we only include those things necessary running on minikube? I think EFS is AWS-specific, right?

scrapyd_k8s.sample-k8s.conf
```diff
@@ -57,3 +57,16 @@ requests_cpu = 80m
...
+
+;[joblogs]
+;# Choose storage provider
+;storage_provider = s3
+;container_name   = <container_name>
+
+;# Choose number of unique logs, but at least 2
+;num_lines_to_check = 2
+
+;[joblogs.s3]
```
wvengen, line 69: Let's name this `joblogs.storage.s3`, to avoid any future collison when there may be a need for more sections.

scrapyd_k8s/config.py
```diff
@@ -21,13 +21,27 @@ def launcher_cls(self):
...
+
     def listprojects(self):
         return self._projects
 
     def project(self, project):
         if project in self._projects:
             return ProjectConfig(self._config, project, self._config['project.' + project])
 
+    def get_namespace(self):
+        return self.scrapyd().get('namespace', 'default')
```
wvengen, line 43: We don't use the prefix `get_` elsewhere, so I think we can just call this method `namespace`.

scrapyd_k8s/config.py
```diff
@@ -21,13 +21,27 @@ def launcher_cls(self):
         pkg, cls = repo.rsplit('.', 1)
         return getattr(import_module(pkg), cls)
 
+    def joblogs(self):
+        if self._config.has_section('joblogs'):
+            return self._config['joblogs']
+        else:
+            return {}
```
wvengen, line 28: This can be written shorter as `self._config.get('joblogs', {}), right?
Or it could also make sense to return `None` if no joblogs section is present, I think that would be clearer perhaps?

scrapyd_k8s/config.py
```diff
@@ -21,13 +21,27 @@ def launcher_cls(self):
...
+
+    def joblogs_provider_args(self, provider):
```
wvengen, line 30: what about `joblogs_storage` for the function name?

scrapyd_k8s/log_handler.py
```diff
@@ -0,0 +1,161 @@
...
+
+watcher_threads = {}
+pod_tmp_mapping = {}
+
+def get_last_n_lines(file_path, num_lines):
+    """Efficiently retrieve the last `num_lines` lines from `file_path`."""
+    try:
+        with open(file_path, 'rb') as f:
+            # Move to the end of the file
+            f.seek(0, os.SEEK_END)
```
wvengen, line 19: Ah, great you're using seek here!

scrapyd_k8s/log_handler.py
```diff
@@ -0,0 +1,161 @@
...
+                    block_data = f.read(file_size)
+                file_size -= block_size
+                data = block_data + data
+                lines_found = data.count(b'\n')
+                # If we've found enough lines, we can stop reading more blocks
+                if lines_found >= remaining_lines:
+                    break
+
+            # Decode the data and split into lines
+            lines = data.decode('utf-8', errors='replace').splitlines()
```
wvengen, line 39: Just to be sure: do we know this is `utf-8`?

scrapyd_k8s/log_handler.py
```diff
@@ -0,0 +1,161 @@
...
+            result = lines[-num_lines:]
+            return result
+    except FileNotFoundError:
+        logger.warning(f"File not found: {file_path}")
+        return []
+
+def concatenate_and_delete_files(main_file_path, temp_file_path):
+    try:
+        with open(main_file_path, 'a') as main_file, open(temp_file_path, 'r') as temp_file:
+            main_file.write(temp_file.read())
```
wvengen, line 50: Would this read the whole file into memory?
If so, please do it by chunks to avoid a memory hog (and greatly increasing the required memory resources to run scrapyd-k8s).

scrapyd_k8s/log_handler.py
```diff
@@ -0,0 +1,161 @@
...
+        os.remove(temp_file_path)
+        logger.info(f"Concatenated '{temp_file_path}' into '{main_file_path}' and deleted temporary file.")
```
wvengen, line 52: I would classify this as debug log.

scrapyd_k8s/object_storage/libcloud_driver.py
```diff
@@ -0,0 +1,79 @@
...
+        for arg, env in args_envs.items():
+            env_value = os.getenv(env)
+            if env_value is None:
+                logger.error(f"Environment variable '{env}' for argument '{arg}' is not set.")
+                raise ValueError(f"Environment variable '{env}' for argument '{arg}' is not set.")
+            args[arg] = env_value
+
+        driver_class = get_driver(self._storage_provider)
+        print("CLOUD CREDENTIALS")
+        print(args)
```
wvengen, line 33: don't forget to cleanup the debugging statements :)

scrapyd_k8s/object_storage/libcloud_driver.py
```diff
@@ -0,0 +1,79 @@
...
+    def upload_file(self, local_path: str):
+        object_name = os.path.basename(local_path)
+        try:
+            container = self.driver.get_container(container_name=self._container_name)
+            self.driver.upload_object(
+                local_path,
+                container,
+                object_name,
+                extra=None,
+                verify_hash=True,
```
wvengen, line 50: Just to be understand: why did you choose to verify the hash here, is that common practice? Does it incur extra costs?

scrapyd_k8s/log_handler.py
```diff
@@ -0,0 +1,161 @@
...
+                            'job_name': pod.metadata.name,
+                            'namespace': namespace,
+                            'num_lines_to_check': num_lines_to_check
+                        }
+                    )
+                    watcher_threads[thread_name].start()
+            elif (pod.status.phase == 'Succeeded' or pod.status.phase == 'Failed') and pod.metadata.labels.get("org.scrapy.job_id"):
+                log_filename = pod_tmp_mapping.get(pod.metadata.name)
+                if log_filename is not None and os.path.isfile(log_filename) and os.path.getsize(log_filename) > 0:
+                    if object_storage_provider.is_local_file_uploaded(log_filename):
```
wvengen, line 139: I find the method name a bit confusing here, `object_exists` / `file_exists` or so would be clearer to me. As it is just checking whether an object is already present in the object storage or not.

scrapyd_k8s/log_handler.py
```diff
@@ -0,0 +1,161 @@
+import os
```
wvengen, line 1: This file being in the project's code root, I would expect it to be backend-agnostic. Now there are many Kubernetes-specific parts, so I would move this to `joblogs/log_handler_k8s.py` or so, to make it clear that this is specific to Kubernetes, and to make room for a Docker-based implementation (which would be possible, as far as I currently understand).

scrapyd_k8s/object_storage/libcloud_driver.py
```diff
@@ -0,0 +1,79 @@
...
+            if env_value is None:
+                logger.error(f"Environment variable '{env}' for argument '{arg}' is not set.")
+                raise ValueError(f"Environment variable '{env}' for argument '{arg}' is not set.")
+            args[arg] = env_value
+
+        driver_class = get_driver(self._storage_provider)
+        print("CLOUD CREDENTIALS")
+        print(args)
+        try:
+            self.driver = driver_class(**args)
```
wvengen, line 35: Really happy that you've been able to generalize the storage driver configuration, not needing to hardcode the possible keys for storage providers anymore :)

scrapyd_k8s/object_storage/libcloud_driver.py
```diff
@@ -0,0 +1,79 @@
...
+            raise ValueError('Container name is not set')
+
+        args_envs = config.joblogs_provider_args(self._storage_provider)
+        args = {}
+        for arg, env in args_envs.items():
+            env_value = os.getenv(env)
+            if env_value is None:
+                logger.error(f"Environment variable '{env}' for argument '{arg}' is not set.")
+                raise ValueError(f"Environment variable '{env}' for argument '{arg}' is not set.")
+            args[arg] = env_value
```
wvengen, line 29: I'm not _quite_ sure what happens here. Is it so that all items listed in the provider args, are expected to be set in the environment? While that would work, I think it would be more usable / easier to use when some items could be set in the configuration file, and some could be gotten from the environment.

A very simple approach would be variable expansion, so `${ABC}` would get the value from the `ABC` environment variable (and the need to allow escaping `$` when needed in the value itself). Or let `region = eu-west-1` set it directly, and `region_env = AWS_REGION` get it from the environment.
_Ideally_ this might be generalizable to the whole configuration file, allowing for more flexibility, but I think that actually would make it too complex (one can still generate the config file if you need it) - but such an approach may still be helpful to think about a generic function that does this with a config section or so.

wvengen: I _would_ be ok with keeping it as it is, provided that we mark this functionality as alpha or so, that users should expect that the configuration approach of this section can still change. Since it would also be nice to converge a bit to getting this merged. (and then work on this later, by you or someone else)

wvengen (commented):
Thanks for the changes. I think it's getting into a good shape, well done! Some questions pending, but nothing that big, I think.

One general thought is that most of the code uses classes, with exception from the http handlers, and utility functions. This is a larger piece of functionality, that, from the outside, I would think would fit the use of a class well.

And thinking more generally about procedural vs. (semi) object-oriented, I think the latter has benefits in code re-use. For example, consider the possibility that scrapyd-k8s would somehow evolve to talk to different clusters at the same time. Then having these global variables with state, would prohibit the current code to be used. When using classes, this state (and setting up connections to k8s etc.) naturally happens within the class instance context, and it would be much easier to get this working. Now this is just a thought exercise, but would nudge me towards preferring the use of a class for job log handling. And I also realize, that when a Docker-implementation would appear, this would naturally be configured by setting a joblog implementation, which would be a class as well.

So if you manage, would you be able to use a class for the Kubernetes joblogs implementation?

vlerkin:
Thank you for your feedback!

> So if you manage, would you be able to use a class for the Kubernetes joblogs implementation?

I do agree with your reasoning and I will work on that, also a great moment for me to learn a bit more about design patterns that may be beneficial for this case.

scrapyd_k8s/config.py
vlerkin: I agree that return None makes more sense in this case

scrapyd_k8s/config.py
vlerkin: Ok with the name if you think it depicts the intention of the function better

scrapyd_k8s/log_handler.py
vlerkin: I had to spend some time to make this new algorithm work:D Reading chunks of bytes is a bit more sophisticated that reading lines in Python:)

scrapyd_k8s/log_handler.py
vlerkin: Isn't utf-8 standard? Even windows made utf-8 sort of default, unless, a person uses some old windows machine to host this cluster, then generated files will probably have their famous typewriter coding. Anyway, I would still expect utf-8.

wvengen: Well, yes, it is kind of the default noawadays, but it may vary in certain cases.
It is just a hardcoded assumption that I'd like to check. In production-grade software, I'd expect one to explain why this is utf-8 (e.g. linking to software generating the log, where it is mentioned - I don't think we know that for sure here), or explain why e.g. it doesn't really matter that much (byte-wise comparison would also be ok, we're just interested in newlines and bytes here I think - with e.g. utf-16 this would matter, but for most other regular encodings, this doesn't make a difference).
Or mention that this is an assumption - I think that would be fine here :)

scrapyd_k8s/log_handler.py
vlerkin: Yes, thank you for catching this.

scrapyd_k8s/object_storage/libcloud_driver.py
vlerkin: It is a good practice because this was we verify that the file has not been corrupted during transmission, if you think it's not critical we can skip it, of course.
It does incur extra cpu costs and for very large files the time can be noticeable.

If we don't care about data integrity here and fast upload is the primary goal, then I'll change it to
```
verify_hash=False
```

wvengen: Thanks for explaning. Let's not enable it by default. It the needs comes up later, we can add it with an extra option (but then we have a whole range of S3-related things we may or may not want to enable, like storage class, acls, perhaps versioning, policies, e.a.).

scrapyd_k8s/object_storage/libcloud_driver.py
vlerkin: I added the alternative, if cloud credentials are not provided as ENV, then it takes values from config, tested for the case you mentioned:
```
    [joblogs.storage.s3]
    key    = ${S3_KEY}
    secret = ${S3_SECRET}
    region = eu-north-1
```
I would discourage any explicit config values to avoid compromising the cloud, but maybe region is not that critical in this case.

wvengen: Super, that sounds great!
There are also other values like endpoint URL, whether to use domain-host-prefixes or not, that may make more sense to put in version control for deployment, vs. secrets. This gives the flexibility for sysadmins to decide that for themselves. :+1:

vlerkin:
I've addressed the comments and also changed the structure to class implementation. The feature contains many methods and lines of code, so I decided to add docstrings to give a short overview on what is going on there. I think it takes less energy to simply read a description and then decide if you want to enable the feature that to read through the entire code.

scrapyd_k8s/object_storage/libcloud_driver.py
```diff
@@ -19,26 +63,73 @@ def __init__(self, config):
...
+            env_value = os.getenv(env_var)
+            if env_value is not None:
+                env_value = env_value.strip().strip('"').strip("'")
+                return env_value
+            else:
+                logger.error(f"Environment variable '{env_var}' is not set for argument '{arg_name}'.")
+                raise ValueError(f"Environment variable '{env_var}' is not set for argument '{arg_name}'.")
+
+        result = self.VARIABLE_PATTERN.sub(replace_var, value)
+        return result
```
wvengen, line 114: Fine!
Did you think about escaping? What if we want to include `${}` in the value (e.g. a password), or `${X}`. I would expect this could be done by `\${}` resp. `\${X}` (or `$${}` resp. `$${X}`, let's pick what makes most sense in the Python world).

scrapyd_k8s/__main__.py
```diff
@@ -13,4 +15,6 @@ def setup_logging():
 
 if __name__ == "__main__":
     setup_logging()
+    config = Config()
```
wvengen, line 18: Does this mean `Config` is instantiated twice, now?

wvengen (commented):
Nice, changes look good!
Great you added docstrings, very helpful. I'm not fully up-to-speed with Python docstring styles, perhaps it's good to check this with others in the team?

scrapyd_k8s/object_storage/libcloud_driver.py
vlerkin: Added a line of code to interpret a common syntax to avoid something and consider it as a part of the password
```
\{}
```

scrapyd_k8s/__main__.py
vlerkin: Yes, sorry, you are right, I can use the instance from api.py file

scrapyd_k8s.sample-k8s.conf
```diff
@@ -57,3 +57,16 @@ requests_cpu = 80m
...
+;storage_provider = s3
+;container_name   = <container_name>
+
+;# Choose number of unique logs, but at least 2
+;num_lines_to_check = 2
+
+;[joblogs.storage.s3]
+;key    = ${S3_KEY}
+;secret = ${S3_SECRET}
+;region = ${S3_REGION}
```
wvengen, line 72: Can use we `# ` to comment this out instead? `;` looks like it's fine, but it's less common to use for commenting out.
Also, let's include a real-world (commented-out) example here, so use a container_name like `scrapyd-k8s-example-bucket`, and explain in a comment with the storage, that one needs to set the `S3_KEY` and `S3_SECRET` environment variables. Please include `region`, so it's clear to readers that one can mix and match between environment and string, also in the storage.s3 section

scrapyd_k8s/api.py
```diff
@@ -148,5 +146,7 @@ def run():
     if config_username is not None and config_password is not None:
         enable_authentication(app, config_username, config_password)
 
+    launcher.enable_joblogs(config)
```
wvengen, line 149: What about putting this in a conditional like  `if config.joblogs() is not None`, I think that makes it more readable, now I think each time I read it: hey, is joblogs always enabled or not?
This would also avoid a warning about joblogs not being enabled when not configured. I think of this feature as an add-on, not as something that everyone will use.

scrapyd_k8s/config.py
```diff
@@ -21,13 +21,27 @@ def launcher_cls(self):
...
+        if self._config.has_section('joblogs'):
+            return self._config['joblogs']
+        else:
+            return None
+
+    def joblogs_storage(self, provider):
+        if not self._config.has_section('joblogs.storage.%s' % provider):
+            return None
+        return self._config['joblogs.storage.%s' % provider]
+
```
wvengen, line 34: Would `self._config.get('joblogs.storage.%s' % provider)` be shorter and do the same? (also `joblogs`)

scrapyd_k8s/joblogs/log_handler_k8s.py
```diff
@@ -0,0 +1,277 @@
...
+        -------
+        list of str
+            A list containing the last `num_lines` lines from the file.
+        """
+        try:
+            with open(file_path, 'rb') as f:
+                # Move to the end of the file
+                f.seek(0, os.SEEK_END)
+                file_size = f.tell()
+                block_size = 6144
```
wvengen, line 103: magic number, please make it a class variable; ideally you'd explain why you've chosen this block size too

scrapyd_k8s/joblogs/log_handler_k8s.py
```diff
@@ -0,0 +1,277 @@
...
+                # Decode the data and split into lines
+                lines = data.decode('utf-8', errors='replace').splitlines()
+                # Return the last `num_lines`
+                result = lines[-num_lines:]
+                return result
+        except FileNotFoundError:
+            logger.warning(f"File not found: {file_path}")
+            return []
+
+    def concatenate_and_delete_files(self, main_file_path, temp_file_path, block_size=6144):
```
wvengen, line 129: same block size here, this really could use a class variable

scrapyd_k8s/launcher/docker.py
```diff
@@ -66,6 +65,10 @@ def cancel(self, project_id, job_id, signal):
             c.kill(signal='SIG' + signal)
         return prevstate
 
+    def enable_joblogs(self, config):
+        # Do nothing, since joblogs is not supported on Docker
+        pass
```
wvengen, line 70: the very least we can do is log an error message here, right?

wvengen (commented):
Super! I think this PR is almost finished.
Ideally, this feature would also be tested on CI. But that is still quite some work, perhaps, requires starting e.g. a minio server, and such. But that would really be helpful. Maybe we can create a new issue for this?

scrapyd_k8s/config.py
vlerkin: Well, the signature of the get method says it is used to retrieve the value of a particular option, not the entire section and requires an option as a second argument, that is why I don't think it is suitable here, since it has a different logic
```
get(section, option, *, raw=False, vars=None, fallback=_UNSET)
```

wvengen: Ah, that makes sense. Sorry, I was just looking at what the code around did and didn't think about values/sections. Fine!

scrapyd_k8s/joblogs/log_handler_k8s.py
vlerkin: This size was chosen to set some sort of balance between the memory we use (we need to keep entire block in memory) and the number of I/O operations. The smaller the blocks, the more I/O operations we have.

On a side note, its size can be adjustable, do you want me to add it to the config so users can make it smaller or larger based on their use cases?

wvengen: It could make sense as a config option. But I'm also ok with a class variable. A number like this needs to be explained in the code / documentation, like you do now in the PR comment.

vlerkin: Ok, no problem, I added it as a class variable, included into a doctoring and also left a comment why this particular value is used

wvengen: Super, thank you!

scrapyd_k8s/object_storage/libcloud_driver.py
```diff
@@ -0,0 +1,188 @@
...
+            return False
+        except ContainerDoesNotExistError:
+            logger.error(f"Container '{self._container_name}' does not exist in the cloud storage.")
+            return False
+        except InvalidContainerNameError:
+            logger.error(f"Invalid container name '{self._container_name}'.")
+            return False
+        except Exception as e:
+            logger.exception(f"An unexpected error occurred while checking for object '{object_name}': {e}")
+            return False
```
leewesleyv, line 188: We can remove 4 of returns by moving the return False to the end of the logic

wvengen:
Thanks for your comments, @leewesleyv!
The use of properties in the config class is a nice idea, which we don't currently do. I think that could perhaps be tackled in a different issue, because I would expect the code to use that consistently on other places too.

I think moving config file handling to the config class would make sense (though I there are also arguments to keep joblogs-related code in the joblogs parts). Then if the joblogs-related code in the config class would grow too much (which I don't think will happen with this PR, but I might be wrong), then it could warrant its own joblog-config class. So I'd say: go ahead with @leewesleyv's suggestions, if you think it's a good idea.

scrapyd_k8s/api.py
```diff
@@ -148,5 +148,11 @@ def run():
     if config_username is not None and config_password is not None:
         enable_authentication(app, config_username, config_password)
 
+    if config.joblogs() is not None:
```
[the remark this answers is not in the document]
vlerkin: Hey Wesley, thanks for your review. I don't have anything against property or decorator pattern, but since we don't use it, it really belongs to another PR about refactoring of the whole project, not my logging feature.

scrapyd_k8s/config.py
```diff
@@ -21,13 +21,27 @@ def launcher_cls(self):
...
+    def joblogs(self):
+        if self._config.has_section('joblogs'):
+            return self._config['joblogs']
+        else:
+            return None
+
+    def joblogs_storage(self, provider):
+        if not self._config.has_section('joblogs.storage.%s' % provider):
+            return None
+        return self._config['joblogs.storage.%s' % provider]
```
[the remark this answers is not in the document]
vlerkin: I will also resolve completed or taste matter remarks

scrapyd_k8s/joblogs/__init__.py
```diff
@@ -0,0 +1,25 @@
...
+    ----------
+    config : Config
+        Configuration object containing settings for job logs and storage.
+
+    Returns
+    -------
+    None
+    """
+    joblogs_config = config.joblogs()
+    if joblogs_config and joblogs_config.get('storage_provider') is not None:
```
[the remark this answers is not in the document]
vlerkin: Logging feature is an additional "package" you can use, it's not default, so cluttering core code components with it not a good way, in my opinion.

scrapyd_k8s/joblogs/log_handler_k8s.py
```diff
@@ -0,0 +1,281 @@
...
+        ----------
+        config : object
+            Configuration object containing settings for job logs and storage.
+        """
+        self.config = config
+        self.watcher_threads = {}
+        self.pod_tmp_mapping = {}
+        self.namespace = config.namespace()
+        self.num_lines_to_check = int(config.joblogs().get('num_lines_to_check', 0))
+        self.object_storage_provider = None
```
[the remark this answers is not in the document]
vlerkin: > Also, what happens if num_lines_to_check cannot be cast to an integer?
In case a user completely ignored documentation and comments in config, the code will raise TypeError and stop working. This is a desireable behavior.

> Some of the attributes can also be defined as class attributes
If a user wants to initialize two instances with different object providers then the code won't work properly, so I prefer to keep essential, variable and mutable parameters as instance attributes, not class since each instance has its own configuration and may vary.

scrapyd_k8s/joblogs/log_handler_k8s.py
```diff
@@ -0,0 +1,281 @@
...
+        """
+        try:
+            with open(file_path, 'rb') as f:
+                # Move to the end of the file
+                f.seek(0, os.SEEK_END)
+                file_size = f.tell()
+                block_size = self.DEFAULT_BLOCK_SIZE
+                data = b''
+                remaining_lines = num_lines
+                while remaining_lines > 0 and file_size > 0:
```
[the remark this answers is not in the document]
vlerkin: There are several comments across the algorithm, otherwise a special description that explains how it works belongs to wiki or something like that. If you have specific lines where you expect a specific comment, let me know, because to me the algorithm is clear.

scrapyd_k8s/joblogs/log_handler_k8s.py
```diff
@@ -0,0 +1,281 @@
...
+                        if log_lines_counter == len(last_n_lines):
+                            log_file.write(line + "\n")
+                            log_file.flush()
+                        elif line == last_n_lines[log_lines_counter]:
+                            log_lines_counter += 1
+                        else:
+                            log_lines_counter = 0
+
+                if len(last_n_lines) > log_lines_counter:
+                    self.concatenate_and_delete_files(log_file_path, temp_file_path)
```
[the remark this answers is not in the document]
vlerkin: It will be caught because the whole thing is wrapped inside try/except. Is this the answer you were looking for?

scrapyd_k8s/joblogs/log_handler_k8s.py
```diff
@@ -0,0 +1,281 @@
...
+        temp_file_path : str
+            Path to the temporary log file to be concatenated.
+        block_size : int, optional
+            Size of blocks to read at a time (default is 6144 bytes).
+
+        Returns
+        -------
+        None
+        """
+        try:
```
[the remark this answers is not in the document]
vlerkin: Good point, I changed it so only specific exceptions are caught

scrapyd_k8s/object_storage/libcloud_driver.py
vlerkin: Good point, thank you, I was adding exceptions sequentially while reading docs and experimenting with different scenarios, so I end up in this ugly piece of code:D Fixed!

vlerkin:
I think all the comments were addressed, is there anything else I can do or we are ready to merge?:)
@wvengen

wvengen:
I think it would be ready to merge - and it's time to try this in production :+1:
@leewesleyv if you see any remaining points, feel free to submit a PR with suggestions!

Issue #23, opened 2024-02-16
Finish daemonstatus endpoint


wvengen:
The endpoint `GET /daemonstatus.json` does not return what is expected from scrapyd. Make it so.
https://scrapyd.readthedocs.io/en/latest/api.html#daemonstatus-json

wvengen:
@vlerkin if you could pick this up before looking at scheduling, that would be great - it is a simple issue that would be good to have fixed.

vlerkin:
Ok, I will take this one next, no problem

vlerkin:
Submitted a PR

PR #32, opened 2024-09-17
refactored /daemonstatus

vlerkin:
refactored /daemonstatus to contain node_name, status and the number of pending, running and finished jobs, tested that it shows the actual load

### DIFF
--- a/scrapyd_k8s/api.py
+++ b/scrapyd_k8s/api.py
@@ -25,7 +25,7 @@ def healthz():
 
 @app.get("/daemonstatus.json")
 def api_daemonstatus():
-    jobs = launcher.listjobs()
+    jobs = list(launcher.listjobs())
     return {
         "node_name": config.scrapyd().get("node_name", launcher.get_node_name()),
         "status": "ok",
--- a/scrapyd_k8s/api.py
+++ b/scrapyd_k8s/api.py
@@ -26,15 +26,12 @@ def healthz():
 @app.get("/daemonstatus.json")
 def api_daemonstatus():
     jobs = launcher.listjobs()
-    total_pending = len([j for j in jobs if j['state'] == 'pending'])
-    total_running = len([j for j in jobs if j['state'] == 'running'])
-    total_finished = len([j for j in jobs if j['state'] == 'finished'])
     return {
         "node_name": config.scrapyd().get("node_name", launcher.get_node_name()),
         "status": "ok",
-        "pending": total_pending,
-        "running": total_running,
-        "finished": total_finished
+        "pending": len([j for j in jobs if j['state'] == 'pending']),
+        "running": len([j for j in jobs if j['state'] == 'running']),
+        "finished": len([j for j in jobs if j['state'] == 'finished'])
     }
 
 @app.post("/schedule.json")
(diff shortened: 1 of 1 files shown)


scrapyd_k8s/api.py
```diff
@@ -25,7 +25,23 @@ def healthz():
 
 @app.get("/daemonstatus.json")
 def api_daemonstatus():
-    return { "status": "ok", "spiders": 0 }
+    node_name = config.scrapyd().get("node_name", launcher.get_node_name())
```
wvengen, line 28: this can be moved to the return statement

scrapyd_k8s/api.py
```diff
@@ -25,7 +25,23 @@ def healthz():
...
+    total_pending = 0
+    total_running = 0
+    total_finished = 0
+    projects = config.listprojects()
+    for project_id in projects:
```
wvengen, line 33: what about `for project_id in config.list_projects():`?

scrapyd_k8s/api.py
```diff
@@ -25,7 +25,23 @@ def healthz():
...
+        jobs = launcher.listjobs(project_id)
```
wvengen, line 34: Why do you iterate over the projects? It would be much lighter to query all jobs in one go, without the project filter.

wvengen (commented):
Ah, great, thank you!
This can be a little more performant and concise. Otherwise, good to go.

scrapyd_k8s/api.py
vlerkin: Yeah, I thought that jobs can be divided across projects and if I won't check all the jobs of all the projects, I might miss something, I guess we can just simply access all the jobs

scrapyd_k8s/api.py
```diff
@@ -25,7 +25,17 @@ def healthz():
 
 @app.get("/daemonstatus.json")
 def api_daemonstatus():
-    return { "status": "ok", "spiders": 0 }
+    jobs = launcher.listjobs()
+    total_pending = len([j for j in jobs if j['state'] == 'pending'])
+    total_running = len([j for j in jobs if j['state'] == 'running'])
+    total_finished = len([j for j in jobs if j['state'] == 'finished'])
```
wvengen, line 31: I think these can also move to the return statement, right?

wvengen (commented):
Thanks! Almost there :)

scrapyd_k8s/api.py
wvengen: It's just a matter of codestyle, terseness vs. smaller bits per line. Thanks for fixing!

wvengen:
Not sure why the test is failing now, but would be good to look into.

wvengen:
Restarted the test, and is fine. Hope this happens too often :+1:

wvengen:
Ah, good one, about the iterator! Didn't realise it either when reading :o:

wvengen:
Ah, please mention the issue number when opening a PR :) #23 here

(And as a sidenote, we could also make a list of just the state, not requiring the whole objects in memory; but I'm not sure if that improves the code (readability), I was happy to merge it as it is.)

PR #35, opened 2024-11-04
28 persistent logs

vlerkin:
Changed object_exists method to check if at least one object with partial name (job id) exists in the object storage instead of matching the whole name (which is problematic due to a random variable part of the file name.


wvengen:
Thanks for explaining :+1:
(it's helpful if changes are not only understandable to those currently working on the code, but also in retrospect by others)

PR #36, opened 2024-11-07
6 scheduling

vlerkin:
What happens in the PR:
1. The logic of event watcher was separated in an observer class; the logic of log watching stayed in a log handler class, but the initialization was changed to subscribe to the event in case jobless feature was configured;
2. The new class KubernetesScheduler was created to handle logic when jobs must be unsuspended and how (ordered);
3. scheduler endpoint was modified, logic to set a value for start_suspended parameter was added;
4. schedule method from k8s launcher has a new start_suspended parameter, it's value is passed when called inside the api; also new methods were added: unsuspend_job patches existing suspended job suspend=False, get_running_jobs_count returns the number of jobs that are currently running, list_suspended_jobs returns the list of jobs where spec.suspend is true, _get_job_name extracts the job name from the metadata, it is then used for unsuspend function;

The big picture:
Event watcher connects to the k8s api and receives the stream of events, it then notifies the subscribers if a new event is received and passes it to the provided callback. The subscriber - KubernetesScheduler - receives event in a handle_pod_event method, this method reacts to the changes in job statuses, and if job completed running or failed it calls another method - check_and_unsuspend_jobs - that checks capacity and unsuspends jobs until the number of allowed parallel jobs is reached, while doing this it relies on another method - get_next_suspended_job_id - to unsuspend the most recent job, to keep the order in which jobs were initially scheduled.
When the job is scheduled, based on the number of currently active jobs and max_proc provided in the config (default is 4), the job runs or goes to the queue of suspended jobs (native k8s queue). Then events that change the number of active jobs trigger the logic of KubernetesScheduler class that unsuspend suspended jobs until the desired state (num of parallel jobs) is achieved.


kubernetes.yaml
```diff
@@ -181,7 +183,7 @@ rules:
     verbs: ["get"]
   - apiGroups: ["batch"]
     resources: ["jobs"]
-    verbs: ["get", "list", "create", "delete"]
+    verbs: ["get", "list", "create", "patch", "update", "delete"]
```
wvengen, line 186: do you need `patch` _and_ `update`? what is the difference?

wvengen (commented):
Ah, nice you were able to come up with something so quickly already!
I looked at it from a high level, and noticed that this is currently implemented for Kubernetes only, and also setup in such a way that it needs refactoring for Docker. I would think of the scheduler as something that could work for both Docker and Kubernetes, especially the scheduling decisions. Also, there is now k8s-specific code in the main file (e.g. the import), and the kubernetes scheduler, this makes the code somewhat spaghetti: there are specific implementation-specific classes where responsibility is meant to be delegated. If you need to access the scheduler in the main file, use a generic scheduler, and make the docker-based parts not implemented. I think that would give a much cleaner design.

Also, I would consider making the launcher responsible for scheduling. And then have the scheduler talk to the launcher to actually start jobs.

I'm not yet sure if we should allow running without the scheduler, or if it would always be active.

wvengen:
Hope my feedback was at an angle that helps you at this stage. In any case, well done, keep it going!

p.s. the CI error looks like it could be cause by Kubernetes-specific things having entered into the main api code, which wouldn't work when running with Docker.

vlerkin:
Working on Docker implementation to be added to this PR

kubernetes.yaml
vlerkin: Yes, good point, since we do not update the entire resource, patch is enough for the feature, deleted it

scrapyd_k8s.sample-docker.conf
```diff
@@ -16,6 +16,9 @@ repository   = scrapyd_k8s.repository.Local
 # Since this is the Docker example, we choose Docker here.
 launcher     = scrapyd_k8s.launcher.Docker
 
+# Maximum number of jobs running in parallel
+max_proc     = 1
+
```
wvengen, line 21: not sure yet if we want to enable this by default
do you know what scrapyd has in its default configuration?

scrapyd_k8s/api.py
```diff
@@ -7,6 +7,7 @@
 from natsort import natsort_keygen, ns
 
 from .config import Config
+from .k8s_resource_watcher import ResourceWatcher
```
wvengen, line 10: reference to k8s-specific term `k8s_resource_watcher`, this file should use the configured backend and launchers instead

scrapyd_k8s/k8s_resource_watcher.py
```diff
@@ -0,0 +1,106 @@
+import threading
+import logging
+import time
+from kubernetes import client, watch
+from typing import Callable, List
+
+logger = logging.getLogger(__name__)
+
+class ResourceWatcher:
```
wvengen, line 9: Very nice, this approach with subscribers!

scrapyd_k8s/k8s_resource_watcher.py
```diff
@@ -0,0 +1,106 @@
...
+        Parameters
+        ----------
+        event : dict
+            The Kubernetes event data.
+        """
+        for subscriber in self.subscribers:
+            try:
+                subscriber(event)
+            except Exception as e:
+                logger.exception(f"Error notifying subscriber {subscriber.__name__}: {e}")
```
wvengen, line 76: why catch this? when do you expect this?

scrapyd_k8s/k8s_scheduler.py
```diff
@@ -0,0 +1,96 @@
+import logging
```
wvengen, line 1: Can we fit this in the directory structure? I wouldn't expect this in the src root.

wvengen: For me, this functionality seems related to the kubernetes launcher.

scrapyd_k8s/launcher/k8s.py
```diff
@@ -121,8 +135,65 @@ def cancel(self, project, job_id, signal):
             )
         return prevstate
 
-    def enable_joblogs(self, config):
-        joblogs_init(config)
+    def enable_joblogs(self, config, resource_watcher):
+        joblogs_init(config, resource_watcher)
+
+    def unsuspend_job(self, job_id: str):
```
wvengen, line 141: is "unsuspend" the k8s terminology? I would also imagine resume, or start or so.

scrapyd_k8s/launcher/k8s.py
```diff
@@ -121,8 +135,65 @@ def cancel(self, project, job_id, signal):
...
+                namespace=self._namespace,
+                label_selector=label_selector
+            )
+            # Filter jobs where spec.suspend == True
+            suspended_jobs = [job for job in jobs.items if job.spec.suspend]
+            logger.debug(f"Found {len(suspended_jobs)} suspended jobs.")
+            return suspended_jobs
+        except Exception as e:
+            logger.exception(f"Error listing suspended jobs: {e}")
+            return []
```
wvengen, line 196: Is there a situation where you would want to get both the running and suspended job count? Then it could be nice to do one call to `list_namespaced_job` to obtain both.

scrapyd_k8s/launcher/k8s.py
```diff
@@ -145,6 +216,14 @@ def _get_job(self, project, job_id):
 
         return job
 
+    def _get_job_name(self, job_id: str):
+        label_selector = f"{self.LABEL_JOB_ID}={job_id}"
+        jobs = self._k8s_batch.list_namespaced_job(namespace=self._namespace, label_selector=label_selector)
+        if not jobs.items:
+            logger.error(f"No job found with job_id={job_id}")
+            return None
+        return jobs.items[0].metadata.name
```
wvengen, line 225: when you're listing jobs, would you also get the name already?

wvengen (commented):
Great to see a working version! Quite readable :)
I think it needs a little cleanup, but you're getting there, I think.

scrapyd_k8s/api.py
vlerkin: It is required for conditional start of joblogs feature in the run method and this method does not rely on a specific launcher, that is why I can't think of a clean way how to remove it from the api.py. But I am open for suggestions if you see a better way.

scrapyd_k8s.sample-docker.conf
vlerkin: In scrapyd the max_proc parameter is not default, so it can be set. I think there are reasons to consider making it default because it enhances cluster stability but on the other hand, if a user did not think of optimal resource usage this batch limiting can lead to a smaller output at a given time. So it's beneficial if we need to be conscious about recourses of the prod setting but not very handy if we want to go all in and extract as much data as possible and faster.

scrapyd_k8s/api.py
wvengen: If the `ResourceWatcher` is not Kubernetes-specific, perhaps the `k8s_` could be dropped from the filename?

scrapyd_k8s/k8s_resource_watcher.py
vlerkin: This exception is added so we can separate the exception originated in particular subscriber from watcher and from other subscribers.
If there is an unexpected edge case for subscriber it is nice to catch it and understand where it comes from, also easier to debug.

scrapyd_k8s/api.py
vlerkin: I do agree that it shouldn't be here and we don't want to violate SOLID, I will think how to refactor this.

scrapyd_k8s.sample-docker.conf
wvengen: Thanks for your thoughts. I think we'd like the default to be the minimum needed, while still having the important elements (e.g. resource limits are important for scheduling). In a cluster there are already limits, so I think it should be fine without. Similarity to scrapyd is also a plus, so I'd favour not having this set by default.

scrapyd_k8s/k8s_resource_watcher.py
leewesleyv: I think the broad `Exception` should be part of this try-except clause since it is unpredictable what callbacks are executed (not really since we only have 2 currently, but in terms of future proofness). Also, one of the callbacks also raises a bare `Exception` when an issue is encountered. We also need to make sure that when one of the subscribers fails, we do send the event to the other subscribers.

However, it might be better to also re-raise that bare `Exception` as a custom exception so we can do more specific error handling and logging. Same goes for the other callback. Maybe we can design a custom exception that can be raised in the callback that we can then catch while executing these callbacks.

scrapyd_k8s/api.py
vlerkin: Resolved as a part of PR #38

scrapyd_k8s/launcher/k8s.py
vlerkin: The native term for k8s is "suspend", this is the particular field that we set to True/False when we create a job, that if why "unsuspend" seems a good fit and quite descriptive for Kubernetes.

scrapyd_k8s/launcher/k8s.py
vlerkin: Yes, good point, working on it

vlerkin: So, I've been thinking about this, but those are two very different flows that require the latest cluster state when called, but I can make the code a bit more DRY.

vlerkin:
I have problems because I separated this PR partially and now I have a multiverse which I need to refactor to the only source of truth. Going to spend some uncertain amount of time on that.

wvengen:
The way I would do this:
1. Continue working on this PR, until you need the functionality developed in the other PR (or until it has been merged).
2. Interactive rebase on the branch of the other PR. Filter out the commits you had here that you rewrote in the other branch.
3. There may be little or much work to do in resolving conflicts. If it is really many, in various commits, you may consider another route (see below).
4. Test, done.

Of this is much work in many commits, you may consider first doing an interactive rebase of _this_ PR, to simplify it, and reduce the number of commits (that each may need amending).

Yes, this is a bit of work, but something I come across now and then, in various projects.
Sorry for the complexity!

vlerkin:
Thank you for the advice!
I was thinking of dropping the commit with the merging main to this branch, then make the code work so the tests run if needed. Then merging with that other branch that refactored the observer further and make the code of both branches work together and then check if there are any conflicts with main and resolving those. This is a bit longer way than simply redoing the merge with the main branch but I messed up the last one because I lost track of changes, so gradually rebuilding this branch is a bit easier for me.

No worries, this is me who messed up merging, complexity is part of the job:D Learning to make more granular commits and cleaner PRs the hard way:D

scrapyd_k8s/k8s_scheduler.py
vlerkin: k8s_scheduler is not the part of launcher because it is 1) a subscriber and should have this functionality to subscribe to the observer, 2) be the part which is initialized optionally when a user wants to limit the number of parallel job (launcher is not optional), 3) contains higher-level logic that uses low-level methods and helper methods from the launcher.

If you don't like that this file is located in the root, I can relocate it to the directory that belongs to this feature, say, limit_jobs or something like this, you can pick any name you like and I will add it.

scrapyd_k8s.sample-docker.conf
vlerkin: Both implementations are now optional and are activated if a user provided max_proc in the scrapyd section, it is also described in details in CONFIG.md

scrapyd_k8s/k8s_resource_watcher.py
vlerkin: Thank you for your input, colleagues!
What I see here is:
1. Keep the broad exception for the Observer since we cannot guarantee what subscribers do and we do not want one subscriber to mess up all the features that rely on the Observer and Observer itself.
2. Think of custom exceptions for subscribers. - this is something I am going to work on.

vlerkin:
I modified one of the methods in the scheduler (get_next_suspended_job_id) to handle cases if a job does not have a creation_timestamp. It is not expected but if someone used a custom resource and forgot to add this field or made any other error, the job will get the timestamp assigned and will be processed like other jobs in the queue.

Also, there are now unit tests that cover different scenarios for the scheduler.

If you have any other comments for improvements, let me know!

scrapyd_k8s/joblogs/log_handler_k8s.py
```diff
@@ -8,12 +8,20 @@
 
 logger = logging.getLogger(__name__)
 
+# Custom Exceptions
```
leewesleyv, line 11: I'd move these custom exceptions to a separate file.

vlerkin: Good point! If I make a file with custom exceptions within joblogs module, is it fine? I just feel like making a general file for custom exceptions in the root directory for this optional feature is not a good idea. What do you think?

wvengen:
Could you please resolve conflicts on `main`, so that I can see what this PR specifically changes?

scrapyd_k8s.sample-k8s.conf
```diff
@@ -19,8 +19,15 @@ namespace    = default
...
-max_proc = 10
+# Number of attempts to reconnect with k8s API to watch events, default is 5
+reconnection_attempts = 5
+
+# Minimum time in seconds to wait before reconnecting to k8s API to watch events, default is 5
+backoff_time = 5
+
+# Coefficient that is multiplied by backoff_time to provide exponential backoff to prevent k8s API from being overwhelmed
+# default is 2, every reconnection attempt will take backoff_time*backoff_coefficient
+backoff_coefficient = 2
```
wvengen, line 30: Do we think the average user needs to configure this? If yes, keep. If no, the documentation would be enough.
Good to remove `max_proc`.

scrapyd_k8s/tests/integration/test_api.py
```diff
@@ -44,7 +44,7 @@ def test_listversions_ok():
     assert_response_ok(response)
 
     json = response.json()
-    assert json['versions'] == AVAIL_VERSIONS
+    assert set(AVAIL_VERSIONS).issubset(set(json['versions']))
```
wvengen, line 47: That is useful. But it does not let us test the precise response anymore, and the order is not tested anymore.
I think I'd rather keep the old code, and make sure we set `AVAIL_VERSIONS` better if it would fail?

scrapyd_k8s/tests/unit/k8s_scheduler/test_k8s_scheduler.py
```diff
@@ -0,0 +1,232 @@
...
+    job.metadata.creation_timestamp = None  # Simulate missing creation_timestamp
+
+    mock_launcher.list_suspended_jobs.return_value = [job]
+
+    with patch('scrapyd_k8s.k8s_scheduler.k8s_scheduler.logger') as mock_logger:
+        job_id = scheduler.get_next_suspended_job_id()
+        assert job_id == 'job1'
+        mock_logger.warning.assert_called_with(
+            f"Job {job} missing 'metadata.creation_timestamp'; assigned max timestamp."
+        )
```
wvengen, line 232: Very nice you're testing this!
Yet ... can we think of an integration test? It's much more reliable if we can test this against Kubernetes (and Docker).

With `max_proc = 0` no job should ever run.
With `max_proc = 1` scheduling two jobs, where one has a sleep, we want to see the last job suspended, and started after the first job has ended.
(etc?)

I think this is not that hard either, except we need to test with different configs, something we don't really do yet.
The current test seems tightly coupled to the implementation. An integration test would allow us to refactor the code, and have the test still tell us if it works or not.

wvengen: Anyway, you this may be for later :)

wvengen (commented):
At first glance, well done! Seems to do the job (though haven't tested it).
An integration test would increase my confidence that it performs well (now I'd have to run it locally to see if it actually works - I think it would, but still I would feel necessitated to do so).

Also, I think the scheduling logic is now implemented in K8s and Docker separately. Would it make sense to have a single piece of code decide when to schedule, and let e.g. the launcher and listener be the interface to K8s/Docker? Haven't thought this fully through, but the question comes up.

Some questions and notes remain, otherwise it's well on the way.

vlerkin:
The launcher is already sort of an interface, since every implementation uses its own launcher. Listener cannot be an interface because event watcher is a native Kubernetes thing, if I understand your comment correctly. So for k8s we still use the watcher to start suspended jobs when at least one job is done or deleted, and for Docker we have a background thread that checks the state of existing containers, we have discussed it previously.

scrapyd_k8s.sample-k8s.conf
vlerkin: max_proc is not present by default, yes, so all scheduled jobs will run immediately. But when provided, then scheduler is activated.

scrapyd_k8s/tests/integration/test_api.py
vlerkin: Since tests modify the state, the idempotency is violated, so it's hard to test the exact response and the test was failing because the expected response does not match reality any more.

scrapyd_k8s/tests/unit/k8s_scheduler/test_k8s_scheduler.py
vlerkin: Normally unit tests that rely on a method signature are enough to confirm the way a method works, integration tests are extras, we can add a ticket to create integration tests that would test different config files.

scrapyd_k8s.sample-k8s.conf
wvengen: Can we leave out the reconnection parameters from the sample config? They are documented, and wouldn't need tweaking when getting started.

wvengen:
Rebased on `main`, adapted integration tests to setup with different configuration files.

wvengen:
I'm not fully happy with the current integration tests, using a shellscript to patch the k8s setup. It might be cleaner to add a YAML file with the desired k8s manifest (in this case, for the role), and let the CI script save the cluster state on cluster setup, and restore it before running a test (`kubectl apply -f` first the pristine cluster state, then the test-specific resource; would even save a scale down as it is part of the pristine cluster state - but not the waiting on it, so perhaps remains useful to keep scale down).
Downside would be that the full role state is still necessary, so any change to scrapyd-k8d's required roles need to be included in the test-specific role manifest as well, so in that respect, a patch is actually more to the point.
So perhaps it _is_ fine as it is, just wanted to share my thoughts here.

vlerkin:
# Wait until the second is finished too
    listjobs_wait(jobid2, 'finished', max_wait=STATIC_SLEEP+MAX_WAIT)
    
Just curious why would you wait until the second job is done? If it was scheduled after the first job is done then the feature works properly, I am not sure what we are testing here with the waiting for the second one.

vlerkin:
I would group the unit tests under 4 classes:
TestKubernetesSchedulerInitialization: test_k8s_scheduler_init, test_k8s_scheduler_init_invalid_max_proc

TestPodEventHandling:
test_handle_pod_event_with_non_dict_event, 
test_handle_pod_event_pod_missing_status, etc

TestJobSuspensionManagement: test_check_and_unsuspend_jobs_with_capacity_and_suspended_jobs, test_check_and_unsuspend_jobs_no_suspended_jobs, etc

TestSuspendedJobSelection: test_get_next_suspended_job_id_with_suspended_jobs, test_get_next_suspended_job_id_no_suspended_jobs, etc

This is much more readable and easier to contribute to add or change tests. What do you think?

vlerkin:
And to make code more compact, we could use @pytest.mark.parametrize to avoid code duplications

wvengen:
Hi :wave: Thanks for your input!

> Just curious why would you wait until the second job is done? If it was scheduled after the first job is done then the feature works properly, I am not sure what we are testing here with the waiting for the second one.

As a general note, the integration tests are, integration tests, and meant to test the system as a whole. Here it means that it is not strictly necessary to wait on the second job, but it is still part of the expected flow, and doesn't hurt to test. These tests are more like what a user would expect when using the system; not handling specific edge cases very isolated, but may include edge cases in the integrated flow.

> TestKubernetesSchedulerInitialization, TestPodEventHandling, TestJobSuspensionManagement, TestSuspendedJobSelection

These tests sound Kubernetes-specific, and are not really about testing a full interaction cycle with scrapyd-k8s using its API only. Therefore they don't really belong in the (current) integration tests, I think.

There are probably specific cases to cover, as you write. Very useful to know about. Maybe we could express them as a full integration tests, that triggers a certain corner-case, and should work in a certain way, regardless of backend (k8s/docker).

If we want to test the _launchers_ (incl. the possible _schedulers_), then we'd need another kind of tests, perhaps docker- and k8s-specific tests, that check how a cluster/node responds to launcher commands, and vice versa. Here we might test the surface API of the launcher (instead of the REST API). This is a kind of test we don't have yet (all tests are now backend-agnostic).

In the early stages of this project, to keep testing work managable, no backend-specific tests were created. There may come a time where this project grows, and needs backend-specific tests, but I see the overhead as a bit too much as of now - as long as we can cover enough ground with the backend-independent integration tests.

>  test_k8s_scheduler_init_invalid_max_proc

This seems like it would be a separate integration test with a config file having an invalid value. I think the API does not expose this, so either the daemon does not start at all, or runs with a reduced feature set. Both are not reported through the API, so cannot be well tested now. So this requires a different kind of testing setup, curently out-of-scope. I think this would belong in a different issue, in revising the testing infrastructure.

Note that I'd really like to keep a distinction between running actual tests, and setting up the daemon under test in an environment.

vlerkin:
> These tests sound Kubernetes-specific, and are not really about testing a full interaction cycle with scrapyd-k8s using its API only. Therefore they don't really belong in the (current) integration tests, I think.

They are unit tests, I just looked at them again and suggested possible improvements, it was probably a bit confusing:)

vlerkin:
I don't really see any improvements for the integration tests, looks good as it is

vlerkin:
Hey Willem, is there anything you would like to add to this PR?

scrapyd_k8s.sample-k8s.conf
wvengen: This is something to look at still.

scrapyd_k8s/k8s_resource_watcher.py
```diff
@@ -4,7 +4,6 @@
 from kubernetes import client, watch
 from typing import Callable, List
 import urllib3
-
```
wvengen, line 7: This empty line can remain, perhaps?

scrapyd_k8s/k8s_scheduler/__init__.py
```diff
@@ -0,0 +1 @@
+from scrapyd_k8s.k8s_scheduler.k8s_scheduler import KubernetesScheduler
```
wvengen, line 1: As to my current understanding, the Kubernetes scheduler is part of the Kubernetes launcher (it has nothing to do with the repository, and it is also not involved in joblogs). So I would expect this to be somehow grouped with the Kubernetes launcher: scheduling is an aspect of launching.

Is that a view that matches yours?
If so, I would expect this to be reflected in the file structure.

scrapyd_k8s/tests/unit/k8s_scheduler/test_k8s_scheduler.py
wvengen: Just to document this for future readers: at the current stage, integration tests are the way to test scrapyd-k8s, and unit test are extras (things that are crucial to test, but hard or clumsily to do in an integration test). I have not reviewed whether this unit test follows that criterium, but we have it, and as such it is useful to test. :+1:

scrapyd_k8s/tests/unit/launcher/test_docker.py
```diff
@@ -0,0 +1,182 @@
+import pytest
```
wvengen, line 1: for the record: same note as the Kubernetes scheduler unit test.

wvengen (commented):
Some small notes, but I think it is mostly ready to merge! :tada:

vlerkin:
I made changes you mentioned, could you, please, check if this is what you expected?

CONFIG.md
```diff
@@ -59,18 +59,28 @@ choose to use them.
 The event watcher establishes a connection to the Kubernetes API and receives a stream of events from it. However, the
 nature of this long-lived connection is unstable; it can be interrupted by network issues, proxies configured to terminate
 long-lived connections, and other factors. For this reason, a mechanism was implemented to re-establish the long-lived
-connection to the Kubernetes API. To achieve this, three parameters were introduced: `reconnection_attempts`,
+connection to the Kubernetes API. To achieve this, three parameters were introduced:
 `backoff_time` and `backoff_coefficient`.
```
wvengen, line 63: this now says "three parameters", but lists only two ;)
I think it makes sense to include the `reconnection_attempts` too here.

scrapyd_k8s.sample-k8s.conf
```diff
@@ -19,8 +19,12 @@ namespace    = default
...
 #pull_secret  = ghcr-registry
 
-# Maximum number of jobs running in parallel
-max_proc = 10
+# Minimum time in seconds to wait before reconnecting to k8s API to watch events, default is 5
+backoff_time = 5
+
+# Coefficient that is multiplied by backoff_time to provide exponential backoff to prevent k8s API from being overwhelmed
+# default is 2, every reconnection attempt will take backoff_time*backoff_coefficient
+backoff_coefficient = 2
```
wvengen, line 27: Thanks for removing `max_proc`. But why did you keep the other ones?

scrapyd_k8s/api.py
```diff
@@ -157,6 +156,5 @@ def run():
     config_password = scrapyd_config.get('password')
     if config_username is not None and config_password is not None:
         enable_authentication(app, config_username, config_password)
-
```
wvengen, line 160: I would keep this empty line, when looking at the method layout.

wvengen (commented):
Almost there!

It's always good to take a fresh look at the PR yourself, after you haven't looked it at for a while. And then you're still bound to miss some things that have become too familiar. Thanks for taking the time to finish the last bits!

CONFIG.md
vlerkin: but we removed reconnection_attempts completely and use infinite reconnection but with a backoff_time and backoff_coefficient to not overwhelm the k8s API, I will edit it to "two parameters"

scrapyd_k8s.sample-k8s.conf
vlerkin: Good point, they have defaults, so no need to keep them

wvengen (approved):
Thank you!

PR #38, opened 2024-11-15
6 scheduling observer extraction

vlerkin:
This PR was extracted from the 6-scheduling branch to encapsulate logic for changes in resource watcher and log handler. A couple of changes was added to make watcher more resilient, also the comment about cleaning api.py from k8s resource watcher was taken into account.

What's new?
1. The watcher was refactored to become an Observer who notifies subscribers [log handler] when another event is received.
2. The observer itself is wrapped in the logic to gracefully reconnect if the connection with k8s API is lost, it contains backoff_time, backoff_coefficient and reconnection_attempts that are configurable via config file to prevent k8s API infinite flooding. A couple of exceptions are added to handle cases that have been already encountered and can be potentially faced.
3. File api.py is refactored to remove resource watcher logic and encapsulate it within k8s launcher, a new method _init_resource_watcher was created as part of K8s class to handle the logic of resource watcher and joblogs feature initialization, also joblogs/__init__.py was refactored and cleaned since the current implementation of  _init_resource_watcher and enable_joblogs is sufficient to do the job.


kubernetes.yaml
```diff
@@ -87,6 +87,11 @@ data:
     launcher     = scrapyd_k8s.launcher.K8s
     
     namespace    = default
+    
+    max_proc     = 2
+    reconnection_attempts = 5
+    backoff_time = 5
+    backoff_coefficient = 2
```
wvengen, line 94: Can we have sensible defaults, and not have to think about this yet?
Then we can add a more detailed configuration documentation to document this. I think ideally, no one would need to touch these (unless you have very different network requirements) - I'd like the default to work for most cases.

scrapyd_k8s.sample-k8s.conf
```diff
@@ -19,6 +19,19 @@ namespace    = default
...
+# Number of attempts to reconnect with k8s API to watch events, default is 5
+reconnection_attempts = 5
+
+# Minimum time in seconds to wait before reconnecting to k8s API to watch events, default is 5
+backoff_time = 5
+
+# Coefficient that is multiplied by backoff_time to provide exponential backoff to prevent k8s API from being overwhelmed
+# default is 2, every reconnection attempt will take backoff_time*backoff_coefficient
+backoff_coefficient = 2
+
```
wvengen, line 34: (see previous comment)

wvengen (commented):
Great! Did a first review pass. Some small (I think) questions.
Very nice, seems well handled.

Would you be able to explain a bit more how the resource_version check works? I'm not familiar with it. If this is "common k8s knowledge", then I need to RTFM, otherwise a comment block in the code would be useful.

kubernetes.yaml
wvengen: I think a section in the README would make sense here. Though might it be time to create a new file explaining the configuration options in more detail? That could give some more freedom to explain them in more detail.

wvengen:
p.s. thank you so much for seperating this PR off #36 :pray:

vlerkin:
About resource versions. If you want to just look at the docs, it is described over here:
https://kubernetes.io/docs/reference/using-api/api-concepts/#resource-versions

Otherwise, I will explain a bit if it's ok.
Every object in k8s has a ```metadata.resourceVersion``` field, when we receive a stream of events from k8s API it is available by default. Throughout the whole lifecycle the resource version of each object is being updated, namely, incremented with every update of the resource. When we leverage the k8s watcher to track objects and suddenly lose connection, we can pass the last resource version we saw to say "hey, give me the stream of events starting with this particular object change", so we don't have to go through the whole stream again and we do not lose the updates.
If resource version is too old, we encounter ```410 Gone``` error, it happens because the retention window is several minutes, if we did not manage to establish a new connection to the k8s API within that window, we handle this case by setting resource version to None and establish connection to receive the available data within the available window.
And yes, a resource version is a common thing in k8s, but it's optional to use.

I hope my explanation is not too lame:D

kubernetes.yaml
vlerkin: There are defaults in the code, but they need to be "tested" to see if it's enough for our prod cluster, I suspect that it might be possible that we need to increase the number of attempts, but I tried to make it resilient in a sense that every time the connection was successfully re-established, the number of attempts sets to provided default number, so we really catch the cases when we have several connection breaks in a row.

wvengen: Super, thanks! Pending:
- [ ] Remove the variables from the sample config files (defaults are meant to be ok, and are documented)
- [ ] Move config info from README.md to CONFIG.md
- [ ] Make sure CONFIG.md is linked to from README.md at a sensible place.

vlerkin: Thank you for the feedback, done!

kubernetes.yaml
```diff
@@ -87,6 +87,8 @@ data:
     launcher     = scrapyd_k8s.launcher.K8s
     
     namespace    = default
+    
+    max_proc     = 2
```
wvengen, line 91: Why is `max_proc` here 2, and in `scrapyd_k8s.sample-k8s.conf` 10?

wvengen:
Thanks! I'll merge and start testing, then we can tie up any loose doc ends as the project progresses.

PR #39, opened 2024-11-20
Adding detailed logs and enabling the debug level to watch every step in prod

vlerkin:
remove limited reconnection attempts; cap backoff time with 15 min; configure logs in a separate file logging_config.py that is imported to the __main__.py to trigger logs configuration before other modules inheret it; add debug logs to the watcher to better monitor each step


scrapyd_k8s/logging_config.py
```diff
@@ -0,0 +1,13 @@
...
+def setup_logging():
+    logging.basicConfig(
+        level=logging.DEBUG,
+        format='%(asctime)s %(name)s [%(levelname)s]: %(message)s',
+        handlers=[
+            logging.StreamHandler(sys.stdout)
+        ]
+    )
+
+setup_logging()
```
wvengen, line 13: Uhm, shall we call this in main nevertheless?

Also, please provide a way to configure the logging level. Debug is not a good default, remember, this is meant to be production-ready code (even if the joblogs feature may not work everywhere yet).

vlerkin: Would you like to have on logger instance for the whole project or have different loggers for each feature/part of the project? I can create an instance right in that config file and then import that instance everywhere (this is what we've done on previous project, it was not too big), but sometimes people create multiple logger instances for each logical part. Depending on it I can refactor the code.

Adding the configurable level.

PS the logger configuration is now called upon importing so we don't have to call it again.

wvengen: Well, let's start with a global log level config for the whole project. Then later we might a way to configure this in more detail.

Ok, I'll leave it to you how to configure the log level. It's just, that in re-usable code, I would call the setup in an init part, so that you can import parts of the code and setup your own logging config. What is logged with what level is the responsibility of the code, but the log level and output format, the responsibility of the main program using (parts of) the code.

But let's try to have some speed, so whatever you think makes sense.

kubernetes.yaml
```diff
@@ -89,6 +89,8 @@ data:
     namespace    = default
     
     max_proc     = 2
+    
+    logging_level = DEBUG
```
wvengen, line 93: not by default!

scrapyd_k8s.sample-k8s.conf
```diff
@@ -22,6 +22,9 @@ namespace    = default
 # Maximum number of jobs running in parallel
 max_proc = 10
 
+# Set the level for logging (e.g., INFO, DEBUG)
+logging_level = DEBUG
```
wvengen, line 26: neither here

scrapyd_k8s/__main__.py
```diff
@@ -1,16 +1,8 @@
...
-            logging.StreamHandler(sys.stdout)
-        ]
-    )
+from .config_loader import config
+from .logging_config import setup_logging
 
 if __name__ == "__main__":
-    setup_logging()
+    logging_level = config.scrapyd().get('logging_level', 'INFO')
+    setup_logging(logging_level)
```
wvengen, line 6: ah, this looks good :+1:

scrapyd_k8s/config_loader.py
```diff
@@ -0,0 +1,3 @@
+from .config import Config
+
+config = Config()
```
wvengen, line 3: I'm a little weary to do it like this: can we make an instance of it in the main file, and pass that on where needed?

scrapyd_k8s/logging_config.py
```diff
@@ -0,0 +1,29 @@
...
+import sys
+
+VALID_LOG_LEVELS = {
+    'CRITICAL': logging.CRITICAL,
+    'ERROR': logging.ERROR,
+    'WARNING': logging.WARNING,
+    'INFO': logging.INFO,
+    'DEBUG': logging.DEBUG,
+    'NOTSET': logging.NOTSET,
+}
```
wvengen, line 11: doesn't Python have something that converts a string to a logging level?

scrapyd_k8s/logging_config.py
```diff
@@ -0,0 +1,29 @@
...
+
+def setup_logging(logging_level):
+    if not logging_level:
+        logging_level = 'INFO'  # Default to INFO if logging_level is None
```
wvengen, line 15: the default could be part of the config

kubernetes.yaml
wvengen: please leave out this line, in normal use, one shouldn't need to touch it (INFO is fine)

scrapyd_k8s/config_loader.py
vlerkin: If resides in the api.py then I have circular imports, that is why I decided that it should be separate to not cause conflicts. I can also create the config instance in the file with the Config class and import it from there

scrapyd_k8s/logging_config.py
vlerkin: logging.getLevelName() - changed the code to use this, it is a bit less convenient but we don't have to store a dict

scrapyd_k8s/config_loader.py
vlerkin: So, we cannot pass the config from api.py to logging_config.py to __main__.py, we can of course move the logging configuring and related things to the api.py but then we are mixing very different things in one file and end up with a bulk of coupled code. Are you ok with keeping the config instance as a separate logic?

wvengen:
I thought this would be an easy change, but I need to dive more into this, to see the bigger picture. Hence I think this shouldn't be rushed too much then. I will try to come back to this on Thu or Fri. Thanks for your quick PR in any case!

wvengen:
Regarding where to configure logging. It must be done before the repository and launcher classes are created, because stuff happens when they are created. Maybe that is not such a good idea, perhaps these classes should only start to actually do something at a later time. But that is how it is now, and I think fine for now.

The global variables are all (or perhaps most) in `api.py`. Can we put the logging configuration there? e.g.

```python
app = Flask(__name__)
config = Config()
setup_logging(level=config.scrapyd().log_level)
repository = ...
```

This still sets up logging when this file is loaded, but a lot happens already here when the file is loaded, that would need refactoring. If we can keep these 'side-effects' at this place, then it is already more readable / understandable.

vlerkin:
Yes, we can put logging to the entry point of the app, no problem, but I would also create a ticket (not urgent) to refactor this to adhere to better practices so we can discuss it as a team and maybe make it cleaner. But since we are hot fixing stuff, I can put everything together for now, no problem. Working on it.

scrapyd_k8s/api.py
```diff
@@ -1,17 +1,22 @@
 #!/usr/bin/env python3
+import logging
+from .config import Config
+config = Config()
+from .logging_config import setup_logging
+logging_level = config.scrapyd().get('logging_level', 'INFO')
```
wvengen, line 6: Can we use `log_level` instead? I have seen that much more often in various projects.

wvengen (commented):
Thanks for the change, appreciated.

scrapyd_k8s/api.py
```diff
@@ -3,8 +3,8 @@
 from .config import Config
 config = Config()
 from .logging_config import setup_logging
-logging_level = config.scrapyd().get('logging_level', 'INFO')
-setup_logging(logging_level)
+log_level = config.scrapyd().get('logging_level', 'INFO')
```
wvengen, line 6: uhm, and also the config option, primarily!

scrapyd_k8s/logging_config.py
```diff
@@ -1,15 +1,15 @@
 import logging
 import sys
 
-def setup_logging(logging_level):
-    if not logging_level:
-        logging_level = 'INFO'  # Default to INFO if logging_level is None
+def setup_logging(log_level):
+    if not log_level:
+        log_level = 'INFO'  # Default to INFO if logging_level is None
```
wvengen, line 6: We already have a default, right? So we can omit it here.

scrapyd_k8s/logging_config.py
```diff
@@ -1,15 +1,15 @@
...
-    level_name = str(logging_level).upper()
+    level_name = str(log_level).upper()
     numeric_level = logging.getLevelName(level_name)
     if not isinstance(numeric_level, int):
         raise ValueError(
-            f"Invalid logging level '{logging_level}'."
+            f"Invalid logging level '{log_level}'."
         )
     logging.basicConfig(
         level=numeric_level,
```
wvengen, line 15: Do you really need a numeric level here?
[`logging.setLevel`](https://docs.python.org/3.8/library/logging.html#logging.Logger.setLevel) accepts both number and string.

wvengen (commented):
Almost there!

scrapyd_k8s/logging_config.py
vlerkin: Right, double check is not needed here, removed it.

scrapyd_k8s/api.py
vlerkin: What do you mean exactly? I don't understand the comment, sorry.

PR #41, opened 2024-11-26
668 joblogs

vlerkin:
Minor changes for the observer:
1. Adjusted timeout_seconds parameter (used to establish long-lived connection to the k8s api) to 300 sec instead of 0 so it resets the connection every 300 sec instead of keeping it alive constantly, this behavior can trigger proxies to close idle connections after a certain timeout they have.
2. Add thread safety for subscribe, unsubscribe and notify_subscribers methods to avoid race conditions.


wvengen:
Great! Can you please rebase this on `main`? Then ready to merge.

wvengen (approved):

Issue #12, opened 2024-01-29
Expose logs


wvengen:
The listjobs endpoint shows a logs path for finished jobs ([as seen here](https://scrapyd.readthedocs.io/en/latest/api.html#listjobs-json)):

```js
{
    "status": "ok",
    // ...
    "finished": [
        {
            "id": "2f16646cfcaf11e1b0090800272a6d06",
            "project": "myproject", "spider": "spider3",
            "start_time": "2012-09-12 10:14:03.594664",
            "end_time": "2012-09-12 10:24:03.594664",
            "log_url": "/logs/myproject/spider3/2f16646cfcaf11e1b0090800272a6d06.log",
            "items_url": "/items/myproject/spider3/2f16646cfcaf11e1b0090800272a6d06.jl"
        }
    ]
}
```

First add serving logs to the API (just plaintext output, the container log messages).
Then add the extra entry to the listjobs endpoint for finished jobs.

wvengen:
After #28 (incl. PRs #31, #42, #48), this feature can be implemented relatively easily. If the file is present in the joblogs dir, then we can serve it from there (e.g. sendfile, and ideally we'd also support range headers); if it is present on object/container storage instead, it could become a redirect to a signed URL, or alternatively proxied.

wvengen:
Note that implementing #53 would be useful, by adding a label to specify compression method, so that the log can be found on container storage in compressed form (without having to try all extensions).

The file should then be served with the proper [`Content-Encoding`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Encoding) for [compression](https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/Compression). This may need specifying the header when storing the log (or proxying).

PR #43, opened 2024-12-18
668 joblogs

vlerkin:
Added logic to remove files:
1. After file was successfully uploaded to the S3-like-storage the file in the volume will be removed;
2. When code receives events from other existing jobs on the cluster it checks if files with its logs are present in the S3-like-storage, if not the code will attempt to collect logs, if the stdout is still available, if the file exists the code now checks if that file also still present in the volume and removes it since it is present in the S3-like-storage.

I think this functionality addresses your concerns.


wvengen:
Thanks! Can you please rebase this on the `main` branch?

Regarding your second point, I don't fully understand this. Does it mean that if a job is found running or finished, and also a log on container storage, it will do nothing? When finished, I think this makes sense, but when running, this is actually an error condition, I would think. Or can you think of cases when a file uploaded to container storage exists, and a job is still found running?

(Also, I was thinking that this will generate a lot of checking with container storage to see if files exists. That is fine. A later improvement could be to add metadata to the job indicating that its logs have been stored (potentially with the full container storage path, so it can be easily located e.g. by #12). But that's for later.)

Issue #28, opened 2024-03-08
More persistent logs


wvengen:
Currently, Docker / Kubernetes logs are used for logging. This is sometimes good enough, but in many situations not. These logs are often truncated at night (and potentially more often when grown to a large size) - especially on Kubernetes - so inspecting errors of long-running jobs can be difficult.

Find a way to keep logs around for at least a bit longer, e.g. for the lifetime of the job (as that could be configurable).
The focus is on Kubernetes, where this is the most pressing issue.

Note that if you have a large, mature Kubernetes cluster, it likely includes components to handle logs. But for smaller clusters, it brings a lot of overhead, and something else is desired.

Either Kubernetes has some hints or so to keep logs around for longer, or logs need to be stored elsewhere (and also removed by some system).

wvengen:
A more complex solution (to setup) could be something like https://kube-logging.dev/ that forwards logs to a central place and can store them on object storage. This looks like it's meant to be run once for the cluster (using daemonsets), so maybe not ideal here.

wvengen:
A simple solution would be that the spider saves logs locally to node-local storage, and pushes them to object storage when the spider is terminated (also when evicted, for example). Then the logs endpoint (#12) needs to have knowledge of it.

wvengen:
Something like [scrapy-logexport](https://github.com/nicholas-mischke/scrapy-logexport) could be a simple solution.
Ideally scrapyd-k8s could recognize this, find the S3 location and credentials, and be able to serve the logs.

wvengen:
Thank you for looking into this!
Storing logs in a persistent volume is an interesting idea.
Some things that arise:
* Can we use the same persistent volume for all pods? This would ease things a lot. If not, we might need a process to move logs from the pod volume to a general volume, requiring more machinary. Also, scrapyd-k8s could mount this volume too and serve logs from there.
* We'd need a method to cleanup old logs.
* Volume storage is more expensive than object storage. Could we store them on object storage instead? Even streaming ... ?
* Jobs may run for longer than a day, so storing the log after the spider finishes may result in lost parts.
* We don't really have big need to have streaming logs, since we can also look at the Kubernetes logs for that. But for scrapyd-k8s as a general solution, it would be something that makes sense to have (for other use-cases than ours).

vlerkin:
Thank you for your thoughts, I will use them as guidance to look deeper into this possible solution, still learning a lot about k8s and can't answer right away:)

[...]

PR #44, opened 2024-12-19
added functionality to remove successfully ploaded files; additionaly…

vlerkin:
Cherry-picked changes (1 last commit from 668-joblogs branch):

Added logic to remove files:

After file was successfully uploaded to the S3-like-storage the file in the volume will be removed;
When code receives events from other existing jobs on the cluster it checks if files with its logs are present in the S3-like-storage, if not the code will attempt to collect logs, if the stdout is still available, if the file exists the code now checks if that file also still present in the volume and removes it since it is present in the S3-like-storage.
I think this functionality addresses your concerns.


wvengen (approved):
Clear, thanks!

wvengen:
Thanks for merging!
Please prefix the PR number with "PR", and include an issue if possible in the commit line as well when merging, and keep the first line short when possible (additional details can be present on later lines). So somethine like this would have been appropriate:

> Remove already uploaded files (#28, PR #44)
>
> after they have been uploaded, and also when old jobs are present on the cluste the files of those jobs already exist on container storage

Otherwise :+1:

vlerkin:
Sorry, got it

Issue #34, opened 2024-10-31
Configure joblogs storage dir


wvengen:
Following #28, I was trying to figure out where to mount the volume for joblogs, and noticed that while the example kubernetes manifest mounts this at `/data` (though commented out), the files generated live in temporary file space https://github.com/q-m/scrapyd-k8s/blob/7975a70a247771449248d04f53aead55d77c22a3/scrapyd_k8s/joblogs/log_handler_k8s.py#L178
According to [`gettempdir()`](https://docs.python.org/3/library/tempfile.html#tempfile.gettempdir), this is configured in the environment.
But I think this needs a dedicated volume, because other temporary files might be generated too, and if these may be used elsewhere, that comes in the way of computing the required storage requirements.
So let's configure this in the config file for joblogs, with a fallback to the temporary directory.

Note that it is possible that the scrapyd-k8s pod is evicted and restarted on another node (e.g. when a node fails, or during upgrades), so to not lose logs, this shouldn't be a host-local path that disappears when switching hosts.

PR #48, opened 2025-01-23
Configure joblogs storage (PR #34)

vlerkin:
adjust k8s yaml to provide a path for logs file generated by the jobogs feature; create an fsGroup to allow writing to the provided directory, adjust PVC manifest; alter cofig.md file by deleting a clause about /tmp dir since we have an appropriate security context now

### DIFF
--- a/CONFIG.md
+++ b/CONFIG.md
@@ -46,7 +46,7 @@ For Kubernetes, it is important to set resource limits.
 TODO: explain how to set limits, with default, project and spider specificity.
 
 ### [joblogs] section
-* `logs_dir`     - a directory to store log files collected on k8s cluster (implemented only for Kubernetes). When configuring, keep in mind that in the Dockerfile the `USER` is set to `nobody` so not all directories are writable, but if you make a child directory under `/tmp` you won't encounter permission problems.
+* `logs_dir`     - a directory to store log files collected on k8s cluster (implemented only for Kubernetes). If you are using a Persistent Volume, keep in mind, that the provided path should be mounted in the deployment manifest.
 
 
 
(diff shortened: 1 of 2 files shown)


kubernetes.yaml
```diff
@@ -106,7 +108,7 @@ data:
     limits_memory = 0.5G
     
     [joblogs]
-    logs_dir = /tmp/scrapyd_k8s_logs
+    logs_dir = /data/scrapyd_k8s_logs
```
wvengen, line 111: let's call this `/data/joblogs` - the PVC is just for scrapyd-k8s, and the feature is called "joblogs"

wvengen: Also, shall we drop the PV below? Or at least comment that this may be needed in case x, but not in case y?

CONFIG.md
```diff
@@ -46,7 +46,7 @@ For Kubernetes, it is important to set resource limits.
 TODO: explain how to set limits, with default, project and spider specificity.
 
 ### [joblogs] section
-* `logs_dir`     - a directory to store log files collected on k8s cluster (implemented only for Kubernetes). When configuring, keep in mind that in the Dockerfile the `USER` is set to `nobody` so not all directories are writable, but if you make a child directory under `/tmp` you won't encounter permission problems.
+* `logs_dir`     - a directory to store log files collected on k8s cluster (implemented only for Kubernetes). If you are using a Persistent Volume, keep in mind, that the provided path should be mounted in the deployment manifest.
```
wvengen, line 49: Great. Maybe also mention that it needs to have read and write permissions, e.g. for Kubernetes by setting the `securityContext`'s `fsGroup` (with a link to the k8s docs, if applicable).

wvengen (commented):
Great!
Some small notes, great to see that it was that simple.

wvengen (approved):

Issue #24, opened 2024-02-27
Test http auth


wvengen:
After PR #21, also test that http authentication works as supposed when configured in the config file (see PR #20).

This requires a bit of thinking, because the current integration tests assume no config file changes, but here we would like both to keep testing that the API is accessible without http auth in the default config, and test that it is not accessible when configured.

PR #49, opened 2025-02-26
Integration tests with multiple configuration files

wvengen:
An attempt.

- [x] Initial CI configuration
- [x] #24 
- [x] Implement `-c` argument for scrapyd-k8s
- [x] Check CI configuration functions
- [x] Check `TEST_` environment variables are passed through in CI config (do they need `export`?)
- [x] Find a solution for the final `kill %1` not working the very last time when running scrapyd-k8s directly (both docker and k8s; one thing to try would be to store the pid in a variable right after running it in the background, and killing the pid directly; or use another signal; may also be related to the note on termination in the next point) --> just needed to `wait` and ignore exit code
- [x] Find a solution for in-cluster scrapyd-k8s not coming up again soon enough in 2nd integration test (a likely cause is that scaling down scrapyd-k8s is taking a lot of time, perhaps we need to handle signals better so that it terminates earlier)
- [x] Make sure tests pass
- [x] ~~Document how integration tests function, with config file adaptation, and how to run these tests easily~~ there is no doc about this, fix later

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -129,10 +129,6 @@ jobs:
       - name: Run tests
         run: |
           # setup for in-cluster k8s
-          TEST_WITH_K8S=1
-          TEST_BASE_URL=`minikube service scrapyd-k8s --url`
-          TEST_MAX_WAIT=60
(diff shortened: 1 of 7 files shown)


wvengen:
@vlerkin this is along the lines of what I thought. It keeps a better separation between the environment setup, and the actual test. And exactly the same test can be used for docker/k8s/in-cluster-k8s.

There are some small but important issues to resolve. Are you able to look into this? I think it needs a bit of Linux knowledge, let me know if you need pointers (or ask around).

wvengen:
Ah, the k8s failure is actually very very simple: the service never comes up, because the health check is behind the auth.

Issue #47, opened 2025-01-03
Put project and spider in joblogs path


wvengen:
In #45 joblogs could be stored on container/object storage. This is currently configured with a bucket name, and the jobid is used for the filename. scrapyd uses project and spider name in the logs path (e.g. [see here](https://scrapyd.readthedocs.io/en/latest/api.html#listjobs-json)), let's do that here as well. Is also useful in practice.

As joblogs is still in beta phase, we can change the path without the need for backward compatibility.

vlerkin:
Just a comment to log research results:

remote storages like object storages do not really use traditional directory structure, they use flat object storage model, it means that if we pass a file with a pass like "myproject/spider1/78391cc0fcaf11e1b0090800272a6d06.log" all the words between / are going to be used as prefixes and when we interact with the remote storage to retrieve the files we should know that they are technically prefixes and not directories.

Going to work on a proof-of-concept further.

vlerkin:
So now when we upload the files to the remote storage, we get a log line like this:
Successfully uploaded 'logs/example/quotes/876673320ed011f0b037367e6cfc1db6.log' to container
This is because we use a new structure of the object name
object_name = f"logs/{project}/{spider}/{job_id}.log"

In the AWS S3 I see virtual directories I can follow to find all the log files for a specific project and spider and the type of the logs/ in my s3 in "folder", so it parses the provided path correctly.

Going to make a PR

PR #51, opened 2025-04-01
47-joblogs-path

vlerkin:
#47 
The new changes affect two methods:

KubernetesJobLogHandler.handle_events and LibcloudObjectStorage.upload_file

handle_events now extract two more labels from the job pod to identify project and spider names and then passes them to the upload_file function.

upload_file uses project name, spider name and extract job_id from a local_path parameter to construct a new object name that is used to create a specific directory like structure in the remote storage.

### DIFF
--- a/kubernetes.yaml
+++ b/kubernetes.yaml
@@ -108,7 +108,7 @@ data:
     limits_memory = 0.5G
     
     [joblogs]
-    logs_dir = /tmp/joblogs
+    logs_dir = /data/joblogs
 ---
 apiVersion: v1
 kind: Secret
--- a/kubernetes.yaml
+++ b/kubernetes.yaml
@@ -99,7 +99,7 @@ data:
     env_config   = spider-example-env
     repository   = ghcr.io/q-m/scrapyd-k8s-spider-example
 
-    # It is strongly recomended to set resource requests and limits on production.
+    # It is strongly recommended to set resource requests and limits on production.
     # They can be overridden on the project and spider level.
     [default.resources]
     requests_cpu = 0.2
@@ -108,7 +108,7 @@ data:
     limits_memory = 0.5G
     
     [joblogs]
-    logs_dir = /data/joblogs
+    logs_dir = /tmp/joblogs
 ---
 apiVersion: v1
 kind: Secret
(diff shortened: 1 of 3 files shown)


kubernetes.yaml
```diff
@@ -108,7 +108,7 @@ data:
     limits_memory = 0.5G
     
     [joblogs]
-    logs_dir = /data/joblogs
+    logs_dir = /tmp/joblogs
```
wvengen, line 111: why this change?
Using `/tmp` by default is dangerous, as a pod restart (or moving to another node) can lose joblogs.

scrapyd_k8s/object_storage/libcloud_driver.py
```diff
@@ -141,6 +147,7 @@ def upload_file(self, local_path):
                 verify_hash=False,
                 headers=None
             )
+            logger.debug("HEY I AN UPLOADING")
```
wvengen, line 150: left over ;)

scrapyd_k8s/joblogs/log_handler_k8s.py
```diff
@@ -290,7 +292,7 @@ def handle_events(self, event):
                                 logger.info(
                                     f"Removed local log file '{log_filename}' since it already exists in storage.")
                         else:
-                            self.object_storage_provider.upload_file(log_filename)
+                            self.object_storage_provider.upload_file(project, spider, log_filename)
```
wvengen, line 295: I think it could make sense to move put the project, spider and log filename together in the log handler. And extract the building of the complete log filename, out into a method. This helps keeping things clean when this would be made more configurable.

wvengen:
p.s. please mention the issue in the PR body, so that it is more easily findable in Github.

kubernetes.yaml
vlerkin: yeah, I forgot to change it, I was using tmp for testing it

scrapyd_k8s/joblogs/log_handler_k8s.py
vlerkin: I don't think that two lines of code should be a separate method, object_name is constructed with all the dependencies but we also use local_path to indicate which file we need to upload to the given (object_name) destination, that will lead to a duplications of code which is not ok, these lines really belong here, not in log_handler. Log handler handles locally stored files and local files paths, it's not its responsibility to handle anything for the uploading logic.

job_id = os.path.basename(local_path).replace('.txt', '')
object_name = f"logs/{project}/{spider}/{job_id}.log"

wvengen: Ok, that does make sense. Thanks for explaining!

wvengen (approved):
Thank you! :sparkles:

Issue #45, opened 2025-01-02
Compression for joblogs


wvengen:
After #28 logs can be stored on container storage.
Allow compression of logs, to reduce storage costs. This will also need to play well with #12.
May consider configuring compression (none / gzip / bzip2 / ...).

Ideally, the compression method can be changed and existing logs can still be found (one way would be to search through container storage, which could potentially be expensive; storing the final location as job metadata may be nicer).

wvengen:
Note that we don't need to compress logs on the node (e.g. on a persistent volume). scrapyd-k8s needs to locate the resume point when the connection watching the log is reset (may happen regularly), and and this takes more cpu when compressed. Compress during (or before) uploading to object storage.

wvengen:
Note that gzip still seems good for both read/write, and brotli is good for read-heavy cases (which are logs) (ok compression factor, fast decompression). gzip is slightly better supported by regular text processing tools, so I'd use that as default.
https://cran.r-project.org/web//packages//brotli/vignettes/benchmarks.html

vlerkin:
What I see is than the compression speed is much lower and they say "which makes Brotli less suitable for on-the-fly compression in http servers or other data streams", so I guess if we compress on a fly, this is not our candidate. I would start with gzip and see how it goes. If we opt for a compression as a separate step before uploading to the remote storage, then Brotli looks like a nice-to-have option!

vlerkin:
In the proof of concept there is a couple of lines of code that compresses a file to a gzip and then uploads to the remote storage as a stream of bytes. We still can see the virtual directories structure in the storage, but now the file has an additional extension .gz, so the actual file can be found as "logs/PROJECT_NAME/SPIDER_NAME/JOB_ID.log.gz"

PR #52, opened 2025-04-01
45-log-compression

vlerkin:
#45 

This PR adds a new class that is responsible for compression of the log files. It reads a log file from disk in chunks and writes to a temporary compressed file on a disk (an alternative would be to use an in-memory buffer which is not suitable for large log files, so in sake of user experience it uses disk). Upload function then uses the compression method to upload a file to the remote storage. Depending on the preferred compression method, a log file gets an appropriate extension because of the mapping that is added as a class attribute to the LIbcloud class.

### DIFF
--- a/CONFIG.md
+++ b/CONFIG.md
@@ -47,7 +47,7 @@ TODO: explain how to set limits, with default, project and spider specificity.
 
 ### [joblogs] section
   * `logs_dir`     - a directory to store log files collected on k8s cluster (implemented only for Kubernetes). If you are using a Persistent Volume, keep in mind, that the provided path should be mounted in the deployment manifest. Read and write permissions should be granted to allow actions with log files in the provided directory, thus a [securityContext](https://kubernetes.io/docs/tasks/configure-pod-container/security-context/) was added to the deployment manifest.
-
+  * `compression_method` - a method to compress log files. Available options are `gzip` `bz2`, `lzma` and `brotli`. If not set, defaults to `gzip`.
 
 
 ### Kubernetes API interaction
--- a/CONFIG.md
+++ b/CONFIG.md
@@ -47,7 +47,7 @@ TODO: explain how to set limits, with default, project and spider specificity.
 
 ### [joblogs] section
   * `logs_dir`     - a directory to store log files collected on k8s cluster (implemented only for Kubernetes). If you are using a Persistent Volume, keep in mind, that the provided path should be mounted in the deployment manifest. Read and write permissions should be granted to allow actions with log files in the provided directory, thus a [securityContext](https://kubernetes.io/docs/tasks/configure-pod-container/security-context/) was added to the deployment manifest.
-  * `compression_method` - a method to compress log files. Available options are `gzip` `bzip2`, `lzma`, `brotli` and `none`. The `none` option configures upload to the remote storage without compression.
+  * `compression_method` - a method to compress log files. Available options are `gzip` `bzip2`, `lzma`, `brotli` and `none`. If no compression_method is provided, it defaults to `none`.
 
 
 ### Kubernetes API interaction
--- a/CONFIG.md
+++ b/CONFIG.md
@@ -47,7 +47,7 @@ TODO: explain how to set limits, with default, project and spider specificity.
 
 ### [joblogs] section
   * `logs_dir`     - a directory to store log files collected on k8s cluster (implemented only for Kubernetes). If you are using a Persistent Volume, keep in mind, that the provided path should be mounted in the deployment manifest. Read and write permissions should be granted to allow actions with log files in the provided directory, thus a [securityContext](https://kubernetes.io/docs/tasks/configure-pod-container/security-context/) was added to the deployment manifest.
-  * `compression_method` - a method to compress log files. Available options are `gzip` `bz2`, `lzma` and `brotli`. If not set, defaults to `gzip`.
+  * `compression_method` - a method to compress log files. Available options are `gzip` `bzip2`, `lzma`, `brotli` and `none`. The `none` option configures upload to the remote storage without compression.
 
 
 ### Kubernetes API interaction
--- a/kubernetes.yaml
+++ b/kubernetes.yaml
@@ -107,6 +107,7 @@ data:
     
     [joblogs]
     logs_dir = /data/joblogs
+    compression_method = none
 ---
 apiVersion: v1
 kind: Secret
--- a/kubernetes.yaml
+++ b/kubernetes.yaml
@@ -107,7 +107,6 @@ data:
     
     [joblogs]
     logs_dir = /data/joblogs
-    compression_method = none
 ---
 apiVersion: v1
 kind: Secret
--- a/scrapyd_k8s/object_storage/libcloud_driver.py
+++ b/scrapyd_k8s/object_storage/libcloud_driver.py
@@ -152,14 +152,13 @@ class LibcloudObjectStorage:
 
         try:
             object_name = f"logs/{project}/{spider}/{job_id}.log"
-            if self.compression_method:
+            if self.compression_method != 'none':
                 try:
                     compression = Compression(self.compression_method)
                     compressed_file_path = compression.compress(local_path)
                     file_to_upload = compressed_file_path
-                    if self.compression_method != 'none':
-                        extension = self.COMPRESSION_EXTENSIONS.get(self.compression_method, self.compression_method)
-                        object_name = f"logs/{project}/{spider}/{job_id}.log.{extension}"
+                    extension = self.COMPRESSION_EXTENSIONS.get(self.compression_method, self.compression_method)
+                    object_name = f"logs/{project}/{spider}/{job_id}.log.{extension}"
                 except Exception as e:
                     logger.error(f"Compression failed, will upload uncompressed file: {e}")
                     # Fallback to uncompressed upload
--- a/scrapyd_k8s/object_storage/libcloud_driver.py
+++ b/scrapyd_k8s/object_storage/libcloud_driver.py
@@ -1,8 +1,6 @@
-import gzip
 import os
 import re
 import logging
-import tempfile
 
 from libcloud.storage.types import (
     ObjectError,
@@ -10,6 +8,7 @@ from libcloud.storage.types import (
     InvalidContainerNameError,
 )
 from libcloud.storage.providers import get_driver
+from scrapyd_k8s.object_storage.log_compressor import Compression
 
 logger = logging.getLogger(__name__)
 logging.basicConfig(level=logging.DEBUG)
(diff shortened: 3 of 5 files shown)


scrapyd_k8s/object_storage/log_compressor.py
```diff
@@ -0,0 +1,91 @@
...
+        """
+        # Create a temporary file for the compressed data
+        with tempfile.NamedTemporaryFile(delete=False, suffix=f'.log.{self.method}') as temp_file:
+            temp_compressed_file = temp_file.name
+
+            try:
+                if self.method == "gzip":
+                    with open(input_file_path, 'rb') as f_in:
+                        with gzip.open(temp_compressed_file, 'wb') as f_out:
+                            while chunk := f_in.read(1024):
```
wvengen, line 64: make this a class constant?

scrapyd_k8s/object_storage/libcloud_driver.py
```diff
@@ -38,6 +39,12 @@ class LibcloudObjectStorage:
     """
 
     VARIABLE_PATTERN = re.compile(r'\$\{([^}]+)}')
+    COMPRESSION_EXTENSIONS = {
+        'gzip': 'gz',
+        'bz2': 'bz2',
```
wvengen, line 44: shall we call this `bzip2`?

wvengen (commented):
Wow, you made it with all the different compression methods, way to go :rocket: :)

One thing I am missing, is not to use compression (compression method `none`, which may be the default, for backward compatibility).

Do we need to update the dependencies for brotli (nice btw to make it optional)?

wvengen:
p.s. opened #53 as a follow-up issue.

vlerkin:
Just to be clear because in the ticket you wrote that gzip should be default.

> One thing I am missing, is not to use compression (compression method none, which may be the default, for backward compatibility).

would you like me to make default as None? So no compression at all?

wvengen:
> Just to be clear because in the ticket you wrote that gzip should be default.

Ah, yes, gzip as default compression method. But I think it would be helpful to keep compression off by default, then we can later transition to gzip if we want to. So yes please, default `none`.

vlerkin:
Sorry, one more question to clarify, when you say

> gzip as default compression method

Do you mean that if the provided by user compression method is not listed as implemented, then instead of raising an exception we would proceed with gzip and write a log that explains why we defaulted to gzip? Is this the behavior you have in mind?

wvengen:
Thanks for asking. I think it makes sense to have default compression method `none` - and with multiple compression methods, I don't think there is another default, so you can ignore that sentence now :) (when choosing a compression method without allowing a user to choose, then we'd best have `gzip`, but that's not the case here - well done).

scrapyd_k8s/object_storage/libcloud_driver.py
vlerkin: Is Python standard library this module is called bz2, I follow the same convention for all other modules, if we want to support synonyms from natural language we would need a mapping for a mapping, which is not the problem if you think it's necessary.

wvengen: Thanks for explaining. `bzip2` is more generally used as name, e.g. we use `gzip` here and not `gz`, which would be the shorter name. Even the [Python docs](https://docs.python.org/3/library/archiving.html) call it `bzip2` in running text. Also compare the `tar` manpage

```
       -j, --bzip2
              Filter the archive through bzip2(1).

       -J, --xz
              Filter the archive through xz(1).

       --lzip Filter the archive through lzip(1).

       --lzma Filter the archive through lzma(1).

       -z, --gzip, --gunzip, --ungzip
              Filter the archive through gzip(1).

       -Z, --compress, --uncompress
              Filter the archive through compress(1).

       --zstd Filter the archive through zstd(1).
```

wvengen: So, yes please, `bzip2` here.

vlerkin: This was what I related to:
` import gzip
import bz2
import lzma`

but yes, not a problem to change the name

CONFIG.md
```diff
@@ -47,7 +47,7 @@ TODO: explain how to set limits, with default, project and spider specificity.
 
 ### [joblogs] section
   * `logs_dir`     - a directory to store log files collected on k8s cluster (implemented only for Kubernetes). If you are using a Persistent Volume, keep in mind, that the provided path should be mounted in the deployment manifest. Read and write permissions should be granted to allow actions with log files in the provided directory, thus a [securityContext](https://kubernetes.io/docs/tasks/configure-pod-container/security-context/) was added to the deployment manifest.
-
+  * `compression_method` - a method to compress log files. Available options are `gzip` `bz2`, `lzma` and `brotli`. If not set, defaults to `gzip`.
```
wvengen, line 50: This one also needs to be updated now that the default is not set.

wvengen (commented):
Looking great, thank you!
Please let me know when you think it's ready to merge :)

Can you please verify that when passing multiple configuration files, it is still possible to reset the `compression_method` to the empty value? If not, then we may need to introduce `none` as an explicit value.

wvengen:
Perhaps having an explicit compression method of `none` is useful, while the empty value would bring up the default (this behaviour may be different, if in the future we would opt for a default compression method).

vlerkin:
This is possible, sure, would you like me to add None as an option for the compression_method? The parameter then always should be provided in the config with a desired value.

wvengen:
That would be more explicit. Then we can use `none` as the default value, so it does not need to be in the config file.
This change distinguishes "whatever the default is" from "no compression" as a configuration value.

wvengen:
So, yes please!

vlerkin:
For now I haven't provided any default if the compression_method is not provided in the config file, but when we extract the compression method (with `"none"` included) there might be a default `None` value. So if you want to extend the method later, you could add something you would consider as default when the parameter is not listed in the config. I hope I understood you correctly, otherwise, I can add something else.

kubernetes.yaml
```diff
@@ -107,6 +107,7 @@ data:
     
     [joblogs]
     logs_dir = /data/joblogs
+    compression_method = none
```
wvengen, line 110: we can leave this out, right?

scrapyd_k8s/object_storage/libcloud_driver.py
```diff
@@ -63,6 +70,9 @@ def __init__(self, config):
             logger.error("Container name is not set in the configuration.")
             raise ValueError("Container name is not set")
 
+        # Reading the compression method from the config and setting default to 'gzip'
+        self.compression_method = config.joblogs().get('compression_method', None)
```
wvengen, line 74: I think we can remove 'setting default to gzip' now.

wvengen: Also: I vote for using `'none'` instead of `None`, as this is one less codepath to make sure is handled well (the distinction between those in the code the follows).

kubernetes.yaml
vlerkin: No, because we have none as a mandatory option

scrapyd_k8s/object_storage/libcloud_driver.py
wvengen: yes! That would also allow leaving out the compression_method in the config, right?

vlerkin: Yes, it will, I just thought you wanted another default:) I will make the changes.

scrapyd_k8s/object_storage/libcloud_driver.py
```diff
@@ -136,22 +146,47 @@ def upload_file(self, project, spider, local_path):
...
+                logger.info(f"Successfully uploaded file '{object_name}' to container '{self._container_name}'.")
         except (ObjectError, ContainerDoesNotExistError, InvalidContainerNameError) as e:
             logger.exception(f"Error uploading the file '{object_name}': {e}")
         except Exception as e:
             logger.exception(f"An unexpected error occurred while uploading '{object_name}': {e}")
+        finally:
+            # Remove temporary file even if upload fails
+            if compressed_file_path and compressed_file_path != local_path and os.path.exists(compressed_file_path):
+                os.remove(compressed_file_path)
+                logger.debug(f"Removed temporary compressed file '{compressed_file_path}'.")
```
wvengen, line 189: Is it the case the `compressed_file_path` is only set when the file is compressed, and stored compressed with a different extension? In that case, we don't need to test for `compressed_file_path != local_path` - we can just check if `compressed_file_path` is set, right? That would make the code more explicit, in my view.

scrapyd_k8s/object_storage/libcloud_driver.py
```diff
@@ -38,6 +39,12 @@ class LibcloudObjectStorage:
     """
 
     VARIABLE_PATTERN = re.compile(r'\$\{([^}]+)}')
+    COMPRESSION_EXTENSIONS = {
+        'gzip': 'gz',
+        'bzip2': 'bz2',
+        'lzma': 'xz',
+        'brotli': 'br'
+    }
```
wvengen, line 48: What about moving this to the `Compressor`, e.g. providing a method to return an extension for a compression method, or even giving a compressed filename for an uncompressed filename? Then we only have to adapt one file when we'd add a compression method.

wvengen (commented):
Great! I think we're almost there, just two suggestions that could make the code clearer.
Also, please review the use of single vs. double quotes, I think it's almost consistent.

scrapyd_k8s/object_storage/libcloud_driver.py
vlerkin: Yes, this makes sense, thanks for noticing

scrapyd_k8s/object_storage/libcloud_driver.py
vlerkin: It is a good point to move to the compression class but not a good idea to move responsibility of naming the final upload destination to the compression class since it is fully the responsibility of the upload method in the libcloud class. Would be better to stick to direct functions of each class. I can add a method to provide an appropriate file extension or something like this and move this mapping to the compression class.

wvengen: Hmmm in my mind, providing the path of the compressed file (based on the uncompressed path) is somewhat related to compression, so it would fit there.

But if you think it belongs where it is now, fine to leave it as it is.

vlerkin: It does already return the compressed file path which is then used by the upload function but we also need an extension type to compose the destination in the remote storage for the upload. That is why it seems good to add a new method to get the appropriate extension instead of, for example, parsing the returned compressed file name or something less transparent and maintainable. I am adding the changes, so you can see the exact thing I am referring here to.

Issue #37, opened 2024-11-08
Implement joblogs for Docker


wvengen:
After #35 job logs can be persistent to container storage on Kubernetes.
Ideally, this would also work on Docker.

PR #55, opened 2025-04-28
Add joblog label to job

### LINKED ISSUE #53: Add joblog label to job
After #45, the log filename (also on container storage) is not unambiguous anymore from the project+spider+jobid, as there is a compression method. This compression configuration may be changed, meaning that when #12 is implemented, the log may not be found anymore after change (or else all the extensions need to be queried). To make this work smoothly in the future, I would propose to add a new label with log location. There we can use file extension to determine compression method.

As joblogs are only implemented on Kubernetes now (see #37), this is currently only relevant for Kubernetes.

vlerkin:
#53 
This PR adds the following changes:

A new private method was added to the KubernetesJobLogHandler class: _add_object_name_label.
This method is called after a log file from a corresponding pod was uploaded to the remote storage. This method adds an extension as a label (always the last extension of the file without a dot, so if a file was compressed, it adds a compression extension, if not, then it adds `log` extension. It also adds a label `org.scrapy.log_file_uploaded=true` to indicate that the file was successfully uploaded.

The CONFIG md has a detailed documentation on how a user can reconstruct the final destination in the remote storage to find a specific log file.

### DIFF
--- a/CONFIG.md
+++ b/CONFIG.md
@@ -46,6 +46,36 @@ For Kubernetes, it is important to set resource limits.
 TODO: explain how to set limits, with default, project and spider specificity.
 
 ### [joblogs] section
+
+The joblogs section is used to configure the job logs feature. It is not enabled by default, but can be activated if you
+choose to use it. The job logs feature allows you to collect logs from the Kubernetes cluster and store them in a specified
+directory. The logs are collected from the pods running on the cluster and can be compressed using a specified method.
+The job logs feature is implemented only for Kubernetes and is not available for Docker.
+
+When the log file was uploaded to the specified storage, the pod is labeled with org.scrapy.log_file_uploaded=true, it also
+contains labels to specify an end location of the log files in the storage.
+Since the label field has limitation in the number of symbols, an end location is split into several labels, each of them
+contains a part of the end location. The labels are named as follows:
+  * org.scrapy.project
+  * org.scrapy.spider
+  * org.scrapy.job_id
+  * org.scrapy.extension
+
+How can you build the end location from these labels? Complete the following line
+  * logs/`org.scrapy.project`/`org.scrapy.spider`/`org.scrapy.job_id`.`org.scrapy.extension` 
+
+**Important**: if you used the `compression_method` parameter, the extension of the log file then have two extensions. The
+first extension is `.log` and the second extension is extracted to the label `org.scrapy.extension`. 
+
+For example, if you used
+`compression_method = gzip`, the log file will be named as 
+  * logs/`org.scrapy.project`/`org.scrapy.spider`/`org.scrapy.job_id`.log.`org.scrapy.extension`
+
+If you haven't specified a compression method, then none was applied and the final location of the log file is
+* logs/`org.scrapy.project`/`org.scrapy.spider`/`org.scrapy.job_id`.`org.scrapy.extension`
+
+The joblogs section contains the following parameters:
+
   * `logs_dir`     - a directory to store log files collected on k8s cluster (implemented only for Kubernetes). If you are using a Persistent Volume, keep in mind, that the provided path should be mounted in the deployment manifest. Read and write permissions should be granted to allow actions with log files in the provided directory, thus a [securityContext](https://kubernetes.io/docs/tasks/configure-pod-container/security-context/) was added to the deployment manifest.
   * `compression_method` - a method to compress log files. Available options are `gzip` `bzip2`, `lzma`, `brotli` and `none`. If no compression_method is provided, it defaults to `none`.
 
--- a/CONFIG.md
+++ b/CONFIG.md
@@ -50,7 +50,17 @@ TODO: explain how to set limits, with default, project and spider specificity.
 The joblogs section is used to configure the job logs feature. It is not enabled by default, but can be activated if you
 choose to use it. The job logs feature allows you to collect logs from the Kubernetes cluster and store them in a specified
 directory. The logs are collected from the pods running on the cluster and can be compressed using a specified method.
+If the log file was successfully uploaded to the remote storage, the pod is labeled with `org.scrapy.log_file_uploaded: true`.
 The job logs feature is implemented only for Kubernetes and is not available for Docker.
+
+The joblogs section contains the following parameters:
+
+  * `logs_dir`     - a directory to store log files collected on k8s cluster (implemented only for Kubernetes). If you are using a Persistent Volume, keep in mind, that the provided path should be mounted in the deployment manifest. Read and write permissions should be granted to allow actions with log files in the provided directory, thus a [securityContext](https://kubernetes.io/docs/tasks/configure-pod-container/security-context/) was added to the deployment manifest.
+  * `compression_method` - a method to compress log files. Available options are `gzip` `bzip2`, `lzma`, `brotli` and `none`. If no compression_method is provided, it defaults to `none`.
+
+If you would like to provide a `compression_method` parameter, please, read the dedicated section below.
+
+### compression for joblogs
 
 When the log file was uploaded to the specified storage, the pod is labeled with org.scrapy.log_file_uploaded=true, it also
 contains labels to specify an end location of the log files in the storage.
@@ -60,25 +70,14 @@ contains a part of the end location. The labels are named as follows:
   * org.scrapy.spider
   * org.scrapy.job_id
   * org.scrapy.extension
+  * org.scrapy.compression
 
 How can you build the end location from these labels? Complete the following line
-  * logs/`org.scrapy.project`/`org.scrapy.spider`/`org.scrapy.job_id`.`org.scrapy.extension` 
+  * logs/`org.scrapy.project`/`org.scrapy.spider`/`org.scrapy.job_id`.`org.scrapy.extension`.`org.scrapy.compression` 
 
-**Important**: if you used the `compression_method` parameter, the extension of the log file then have two extensions. The
-first extension is `.log` and the second extension is extracted to the label `org.scrapy.extension`. 
-
-For example, if you used
-`compression_method = gzip`, the log file will be named as 
-  * logs/`org.scrapy.project`/`org.scrapy.spider`/`org.scrapy.job_id`.log.`org.scrapy.extension`
-
-If you haven't specified a compression method, then none was applied and the final location of the log file is
-* logs/`org.scrapy.project`/`org.scrapy.spider`/`org.scrapy.job_id`.`org.scrapy.extension`
-
-The joblogs section contains the following parameters:
-
-  * `logs_dir`     - a directory to store log files collected on k8s cluster (implemented only for Kubernetes). If you are using a Persistent Volume, keep in mind, that the provided path should be mounted in the deployment manifest. Read and write permissions should be granted to allow actions with log files in the provided directory, thus a [securityContext](https://kubernetes.io/docs/tasks/configure-pod-container/security-context/) was added to the deployment manifest.
-  * `compression_method` - a method to compress log files. Available options are `gzip` `bzip2`, `lzma`, `brotli` and `none`. If no compression_method is provided, it defaults to `none`.
-
+**Important**: if you did not use the `compression_method` parameter, the label `org.scrapy.compression` has `none` value
+so the final destination of the log file will be:
+  * logs/`org.scrapy.project`/`org.scrapy.spider`/`org.scrapy.job_id`.`org.scrapy.extension`
 
 ### Kubernetes API interaction
 
--- a/kubernetes.yaml
+++ b/kubernetes.yaml
@@ -166,6 +166,7 @@ metadata:
 rules:
   - apiGroups: [""]
     resources: ["pods"]
+    # add "patch" if you use joblogs feature
     verbs: ["get", "list", "watch"]
   - apiGroups: [""]
     resources: ["pods/exec"]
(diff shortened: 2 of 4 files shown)


CONFIG.md
```diff
@@ -46,6 +46,36 @@ For Kubernetes, it is important to set resource limits.
 TODO: explain how to set limits, with default, project and spider specificity.
 
 ### [joblogs] section
+
+The joblogs section is used to configure the job logs feature. It is not enabled by default, but can be activated if you
+choose to use it. The job logs feature allows you to collect logs from the Kubernetes cluster and store them in a specified
+directory. The logs are collected from the pods running on the cluster and can be compressed using a specified method.
+The job logs feature is implemented only for Kubernetes and is not available for Docker.
```
wvengen, line 54: Let's reference #37 here.

CONFIG.md
```diff
@@ -46,6 +46,36 @@ For Kubernetes, it is important to set resource limits.
...
+Since the label field has limitation in the number of symbols, an end location is split into several labels, each of them
+contains a part of the end location. The labels are named as follows:
+  * org.scrapy.project
+  * org.scrapy.spider
+  * org.scrapy.job_id
+  * org.scrapy.extension
+
+How can you build the end location from these labels? Complete the following line
+  * logs/`org.scrapy.project`/`org.scrapy.spider`/`org.scrapy.job_id`.`org.scrapy.extension` 
+
```
wvengen: Is the extension something that can change right now? If not, then we can leave it out. If the extension would be possible change in the future, we can add this label, and have a default value of `.log`.

CONFIG.md
```diff
@@ -46,6 +46,36 @@ For Kubernetes, it is important to set resource limits.
...
+**Important**: if you used the `compression_method` parameter, the extension of the log file then have two extensions. The
+first extension is `.log` and the second extension is extracted to the label `org.scrapy.extension`.
```
wvengen: The main motivation for the current issue, is that when the config parameter `compression_method` is changed, you cannot find the logs anymore without iterating files. Now including the compression extension solves this, which is nice. But it doesn't tell us what the compression is that was used for the logfile - you still need to analyse the extension and derive the compression method.

So I would rather add a `compression_method` label, and derive the extension. Or, include both the extension and the compression method (so that an external consumer of these labels doesn't need to maintain a mapping between compression method and extension - though it does need to retain a mapping between compression method and decompressor, already, so perhaps this is not an extra burden).

CONFIG.md
```diff
@@ -46,6 +46,36 @@ For Kubernetes, it is important to set resource limits.
...
+
+How can you build the end location from these labels? Complete the following line
+  * logs/`org.scrapy.project`/`org.scrapy.spider`/`org.scrapy.job_id`.`org.scrapy.extension` 
+
+**Important**: if you used the `compression_method` parameter, the extension of the log file then have two extensions. The
+first extension is `.log` and the second extension is extracted to the label `org.scrapy.extension`. 
+
+For example, if you used
+`compression_method = gzip`, the log file will be named as 
+  * logs/`org.scrapy.project`/`org.scrapy.spider`/`org.scrapy.job_id`.log.`org.scrapy.extension`
```
wvengen: this reads like it is an error - why is `log` here?

CONFIG.md
```diff
@@ -46,6 +46,36 @@ For Kubernetes, it is important to set resource limits.
...
+
+If you haven't specified a compression method, then none was applied and the final location of the log file is
+* logs/`org.scrapy.project`/`org.scrapy.spider`/`org.scrapy.job_id`.`org.scrapy.extension`
+
+The joblogs section contains the following parameters:
```
wvengen: This would be a new section.
But I'd rather suggest to move the text explaining to how find the log file on container storage, to its own section, later in the file (not important for users getting started with joblogs, only for those wanting to read them back from their own software), and leaving this joblogs (intro) section as one.

wvengen (commented):
Thanks, also for including documentation! Some small comments.

CONFIG.md
vlerkin: the extension can be either `.log` or `.log.gz` (for example), depending on that, we either extract `log` or `gz`, both scenarios are described in the documentation so we cannot really just leave it out, otherwise it won't be clear what is happening here

CONFIG.md
vlerkin: If I follow your logic, it would be better to have two labels: extension which is `.log` and compression which is either one of the methods or `none`, right?

CONFIG.md
vlerkin: because we extract the last extension, so in a scenario where the compression method was used, the extension would be an appropriate compression method but also `.log`, in a case when a user does not use a compression method the extracted extension is `.log`. If we change the logic following the prev comment then it would be different.

CONFIG.md
wvengen: Well, we can in theory drop the extension, as when there is compression method, we can derive the full extension, right?

p.s. you could include both extensions in the extension, if you really like to have an extension label, e.g. `.log.gz`. But in the comment I explained why I think we don't need the extension, actually.

wvengen: (previous comment deleted) - yes!

wvengen: and I don't think we'd need the extension label then

CONFIG.md
wvengen: Thanks for explaining!

CONFIG.md
vlerkin: why we wouldn't need the extension label? how then a person should assume the final destination without knowing if the file was or was not compressed? I mean it is possible to check the config but since we may rollout update the managing pod and put a new config that has a different compression method or do not use compression at all, it would be pretty hard to trace back the extension of the file. I can entirely drop it if you like, just concerned if it makes the life of the user harder since we can reconfigure things without killing the jobs that sit on the cluster in the completed state.

it is not possible to provide extension as `.log.gz` because the dot `.` is not allowed in labels according to the k8s documentation (mentioned in the ticket)

wvengen: Well, if you have compression method, you can derive the extension, right?

wvengen: Good point about the dot not being allowed. I think that would rule out `s3://` etc URLs (see issue comment).

vlerkin: Well, if you have compression method, you can derive the extension, right? - if it is not in a label then not since we can change a compression method and all previous files then had one compression extension and all future files will have a different compression extension. So if we are documenting the entire destination of the file in the labels, it would be inconsistent to force users to assume a part of it?

wvengen: Well, compression extensions have been stable for many years, this is not something that is likely to change. A new one may be added in the future, but then the code needs to be updated already to support another compression method.

Perhaps ideally, a full URL would be included, but that is hard with the label length and character restrictions. So we'd need to build it anyway from components. Then adding `.log` ourselves, and the compression extension if used, is not such big step anymore.

vlerkin: I am not saying that it's not stable. You run a managing pod with a specific compression method in the config file, the logs are collected to files and shipped in a compressed state. Then you decide not to use any compression and now the cofig does not have any. How can you as a user assume the extension of the existing files to request a particular one?

wvengen: Ah, great we agree on the stability of compression extensions!
You describe how a change in config can result in a different final filename on container storage.
If the compression method is present in the label of a job, you know that job's compression method, right? And then you also know the file's extension, right? Or am I missing something?

vlerkin: But you don't want to include the extension label as I understand from this comment:
> and I don't think we'd need the extension label then

wvengen: Ah, yes! What I (tried to) propose(d) was to replace the label for extension by a label for the compression method. Does that make sense?

vlerkin: I would still keep the extension instead of a method since a user needs to always convert a method to a particular extension and I am not sure what is the sense here. So my question is: do we need to have two extensions in the labels or just one which corresponds to a compressed file extension or to none (then in the path we just write ...`.log.`$COMPRESSED_EXTENSION)?

wvengen: Well, if you put the extension in a label, you still need to derive the compression method from the extension to decompress and read the log. I think it is clearer to put the compression method in the label, and derive the extension (just as the code does) than the reverse.

vlerkin: It is not convenient to return a compression method since it is a separate class and we have a compression extension, it would force us to have a mapping in a log handler class which is not its responsibility, I would still vore for putting the extension in the label and not the method. For the user converting one to another or vice versa is the same amount of mental load. Since we already have the final destination it would be cleaner to use the given info. But if it's important I can add a mapping there.

vlerkin:
I edited some logic and changed the config, if you have something else in mind I can adjust it

CONFIG.md
vlerkin: There is a new section in the config now, could you please check if it looks as you expected?

CONFIG.md
vlerkin: For now I add two labels: extension and compression (compression extension, not method) but as I mentioned if this is something must be (method instead of the extension) then we can add some code, although I do not think it is a clean code

wvengen: I now I understand that my request for adding a compression method label instead of extension, would make the code more complex. I need to look further into this to understand it better, and see if there could be another way. Thanks for explaining!

vlerkin: If you are concerned that we won't be able to decompress a file with "a quasi random compression extension", there are python packages like https://github.com/pycompression/xopen that automatically assume the compression method from the extension of the file and decompress it.

vlerkin: If there are other technical concerns, feel free to share, I can search for possible solutions.

CONFIG.md
```diff
@@ -46,9 +46,38 @@ For Kubernetes, it is important to set resource limits.
...
+
+When the log file was uploaded to the specified storage, the pod is labeled with org.scrapy.log_file_uploaded=true, it also
+contains labels to specify an end location of the log files in the storage.
+Since the label field has limitation in the number of symbols, an end location is split into several labels, each of them
+contains a part of the end location. The labels are named as follows:
+  * org.scrapy.project
+  * org.scrapy.spider
+  * org.scrapy.job_id
+  * org.scrapy.extension
+  * org.scrapy.compression
```
wvengen, line 73: Shall we call this `compression_method`, so it is the same as the key used in the scrapyd-k8s config?

kubernetes.yaml
```diff
@@ -166,6 +166,7 @@ metadata:
 rules:
   - apiGroups: [""]
     resources: ["pods"]
+    # add "patch" if you use joblogs feature
```
wvengen, line 169: let's make it `add "patch" to verbs if you use the joblogs feature` (to make it even clearer that it is about the next line)

CONFIG.md
```diff
@@ -46,9 +46,38 @@ For Kubernetes, it is important to set resource limits.
...
+
+How can you build the end location from these labels? Complete the following line
+  * logs/`org.scrapy.project`/`org.scrapy.spider`/`org.scrapy.job_id`.`org.scrapy.extension`.`org.scrapy.compression`
```
wvengen, line 76: Isn't `extension` always `log`?

scrapyd_k8s/joblogs/log_handler_k8s.py
```diff
@@ -84,6 +84,58 @@ def __init__(self, config):
...
+            elif ".log." in basename:
+                extension = "log"
+                compression = basename.rsplit(".log.", 1)[1]
+
+            patch_body = {
+                "metadata": {
+                    "labels": {
+                        "org.scrapy.extension": extension,
+                        "org.scrapy.compression": compression,
+                        "org.scrapy.log_file_uploaded": "true"
```
wvengen, line 120: If there could be multiple uses of object storage, I propose include `joblogs` in the labels:
`org.scrapy.joblogs.compression_method`
`org.scrapy.joblogs.uploaded`

wvengen (commented):
Thanks for your work on this! Seems mostly fine.
There are some things I would consider changing from an architectural point of view (going beyond this PR, but within the scope of joblogs).

1. I'm confused by `LibcloudObjectStorage` referencing joblogs. If it is joblogs-specific, I would expect it being in the `scrapyd_k8s/joblogs` directory. So there is a decision to make: consider this to be part of joblogs, and move it; or consider it to be really just object storage, and move out all joblog references.
I think that object storage will be used in other places too, in the future (for example to store the items on container storage), so I would tend to refactor out the joblogs references.
Just passing `config.joblogs()` instead of `config` to `LibcloudObjectStorage`'s constructor could be an idea.

2. The compression method from the configuration is kind of the 'input' of the compression, and the path on the object storage is derived from that. That is a reason why I'd prefer storing the compression method as the label, and let the path be derived from it, leaving out the extension. The compression_method value can be taken from `LibcloudObjectStorage` if the approach as described in 1. is taken.

3. Another consideration to take into account in the design: in #12 we will want to access the logs from within scrapyd-k8s. Here we just have the job resource in Kubernetes, and from that we'd like to figure out how to obtain the logs. This would check the `(log_file_)uploaded` label an either look at local or object storage. Here, the job labels need to be consulted, especially regarding compression (because scrapyd-k8s's compression method may have changed from the time this log was stored). I would expect code of storing the logs to be re-used here, especially handling the compression method (though it would need additional methods for reading). Just something to keep in the back of our heads.

vlerkin:
Hey W could you elaborate on joblogs reference in Libcloud class? The only one I see is a section joblogs in the config file, is this what you meant by cleaning out the joblogs from the class?

wvengen:
Hi, thanks for looking into my remarks :) These are joblogs references I could find in the libcloud class:
- config section
- getting the job id from the log filename
- constructing the object name

Issue #40, opened 2024-11-22
Add job `end_time` for Docker in listjobs endpoint


wvengen:
In PR #30, job start end end time was added to the listjobs endpoint. But end_time was only implemented for Kubernetes. This issue is about implementing it for Docker too. This is luckily documented in the README under the [listjobs.json endpoint](https://github.com/q-m/scrapyd-k8s#listjobsjson-).

PR #56, opened 2026-02-17
fix: set end_time on finished job

jrsitabi:
Related to #40 

If a docker containers finishes, its attributes will contain a field FinishedAt.

When /listjobs.json is called, the function api_listjobs() in api.py is called. This function in turn will call the listjobs() function in docker.py. In this function a docker client is called to list all processes. The docker library function does the same as ```docker ps -a```. Then each of these jobs are parsed by the _parse_job() function. Here certain labels and attributes are grabbed and set. Since the FinishedAt field is available, we can grab this and use it to set the end_time field.

To test the functionality the github workflows setup the environment and run tests when pushed to main or a pr is created. The test-docker part succeeds, but the manifest and k8 part fail on the last test ```test_scenario_cancel_running_finished_ok```. The test-manifest step fails somewhere unrelated to to changed files, so more research is needed as to why the assert fails. The test-k8s does fail on a related part, ```assert datetime.strptime(end_time, '%Y-%m-%d %H:%M:%S.%f')```. This logic however is unchanged, so still unclear what the reason for this is. Previously the assert was ```assert end_time is None```, however I suspect that this assert was wrong. After a job is finished the end_time should be available. The assert probably fails, because the line ``` 'end_time': format_datetime_object(job.status.completion_time) if job.status.completion_time and state == 'finished' else None,
``` in k8s.py. The end time doesn't just get set if the state is set to finished, but also if ```job.status.completion_time.```

### DIFF
--- a/scrapyd_k8s/tests/integration/test_api.py
+++ b/scrapyd_k8s/tests/integration/test_api.py
@@ -161,7 +161,7 @@ def test_scenario_cancel_running_finished_ok():
     jobinfo = assert_listjobs(finished=jobid)
     start_time, end_time = jobinfo.pop('start_time'), jobinfo.pop('end_time')
     assert datetime.strptime(start_time, '%Y-%m-%d %H:%M:%S.%f')
-    assert end_time is None
+    assert datetime.strptime(end_time, '%Y-%m-%d %H:%M:%S.%f')
     assert jobinfo == { 'id': jobid, 'project': RUN_PROJECT, 'spider': RUN_SPIDER, 'state': 'finished' }
     # then cancel it again, though nothing would happen
     response = requests.post(BASE_URL + '/cancel.json', data={ 'project': RUN_PROJECT, 'job': jobid })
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: scrapyd_k8s/tests/integration/test_api.py [hunk 1]
```diff
@@ -161,7 +161,7 @@ def test_scenario_cancel_running_finished_ok():
     jobinfo = assert_listjobs(finished=jobid)
     start_time, end_time = jobinfo.pop('start_time'), jobinfo.pop('end_time')
     assert datetime.strptime(start_time, '%Y-%m-%d %H:%M:%S.%f')
-    assert end_time is None
+    assert datetime.strptime(end_time, '%Y-%m-%d %H:%M:%S.%f')
```
wvengen, line 164: This is the job cancel scenario: it seems here that for Kubernetes, the job end time is _not_ available when the job is canceled. For Docker, it seems that it _is_ available. So this is a behaviour difference between the two, and that is not desired.

Can you look at [scrapyd](https://scrapyd.readthedocs.io/) to see what it emits? We want to mirror that behaviour (as far as possible).

wvengen (changes requested):
Well done!

One question to look into: the failing test does matter here.

